PR #6, opened 2016-09-14
Add Python Trace gRPC client

geigerj:


generated/python/grpc-google-devtools-cloudtrace-v1/setup.py
```diff
@@ -0,0 +1,39 @@
...
+"""
+
+import setuptools
+
+from setuptools import setup, find_packages
+
+install_requires = [
+  'oauth2client>=1.4.11',
+  'grpcio>=1.0.0',
+  'googleapis-common-protos[grpc]>=1.3.2'
```
bjwatson: What happened to the upper bounds on `grpcio` and `googleapis-common-protos`? Maybe this was generated before my `packman` PR was merged?

Fixing this is a P2, since we won't probably won't have 2.0 of these packages anytime soon.

PR #3, opened 2016-09-13
WIP: Namespacing changes for Python APIs 0.9.0

bjwatson:
Updates https://github.com/GoogleCloudPlatform/google-cloud-python/issues/2258

PR #15, opened 2016-09-21
Namespacing changes for Python APIs 0.9.x

bjwatson:
Supersedes https://github.com/googleapis/api-client-staging/pull/3.

Note that I do _NOT_ plan to squash this merge, since it captures a sequence of steps:

1) Auto-gen
2) Manual re-gen of ML gRPC packages with custom namespace (which needs to be automated)
3) Hand-edits to gRPC and GAPIC packages (which need to be automated)

The functional test could be broken out into its own PR if necessary, but its easiest just to commit it here and get it into master. There's only one for Vision so far, but we can throw ones together for the other APIs in a separate PR.


bjwatson:
@geigerj Note that I removed the oauth2client upper bound and published `0.9.3` of all packages. This is because we have inconsistent lower-bounds and no upper-bounds across other packages that `google-cloud` depends upon.

PR #18, opened 2016-09-22
Trace ruby

geigerj:
Note: following the style of #15, I intend to merge this as two commits to clearly differentiate generated code from hand-edits in the history.


generated/ruby/google-cloud-ruby/google-cloud-trace/google-cloud-trace-v1.gemspec
```diff
@@ -2,7 +2,7 @@
 # encoding: utf-8
 
 Gem::Specification.new do |s|
-  s.name          = 'google-cloud-trace-v1'
+  s.name          = 'stackdriver-trace-v1'
```
swcloud: I think the gem name should still be google-cloud-trace-v1, just description should be 'Stackdriver Trace'.

bjwatson: Agreed. We're not changing package names in Python in this manner.

swcloud: Close this PR and will open a new one to regenerate

PR #19, opened 2016-09-22
Grpc library for IAM mixins for ruby.

Landrito:


generated/ruby/grpc-google-iam-v1/grpc-google-iam-v1.gemspec
```diff
@@ -0,0 +1,26 @@
+# -*- ruby -*-
+# encoding: utf-8
+
+Gem::Specification.new do |s|
+  s.name          = 'grpc-google-iam-v1'
+  s.version       = '0.6.8'
```
jmuk: Could you update the version to `0.1.0`?  Or `1.0.0` might be even okay, because `grpc` is already `1.0` and `googleapis-common-protos` package is already `1.3.1`.

Landrito: Switched to 1.0.0.

bjwatson: Let's use the 0.x versions that we're currently using for the individual languages. We don't want to declare 1.0 before we have to.

bjwatson: https://github.com/googleapis/packman/blob/master/config/api_defaults.yml#L24

jmuk: Hm, I see, thanks. So `googleapis-common-protos` would be really exceptional it exceeds `1.3`.

bjwatson: Once we're 1.0, it's embarrassing to push breaking changes, as I did a couple weeks ago with Python `googleapis-common-protos 1.3.x` while adding LRO gRPC code.

Landrito: Switched to 0.1.0

bjwatson: There's a separate discussion about whether Ruby should be reset to 0.1.0, or continue down the path it's already been on

bjwatson: https://github.com/googleapis/packman/pull/130

jmuk: Okay, then I approve the patch as-is. You could submit as 0.1.0, or you could make this back to 0.6.8 or something.

Landrito: Okay I'll submit at 0.6.8

PR #20, opened 2016-09-22
IAM Client Library for Python

Landrito:


generated/python/gapic-google-cloud-iam-v1/README.rst
```diff
@@ -0,0 +1,78 @@
+GAPIC library for the Google Admin API
```
bjwatson: Hmm... this should be hand-edited to say "Google IAM Admin API". We have a [Github issue](https://github.com/googleapis/toolkit/issues/270) to improve where we source this title (e.g. from https://github.com/googleapis/googleapis/pull/151/files#diff-b8750c321eaa4d100c91edbf078d08a3R7).

Landrito: Ah I see. Fixed for this case.

generated/python/gapic-google-cloud-iam-v1/docs/conf.py
```diff
@@ -0,0 +1,318 @@
+# -*- coding: utf-8 -*-
+#
+# gax-google-iam-admin-v1 documentation build configuration file
```
Landrito: Should be fixed with the re-gen

bjwatson:
LGTM

bjwatson:
@Landrito Please note that I generated a `0.10.0` version of the gRPC IAM Admin dependency in https://github.com/googleapis/api-client-staging/pull/24.

PR #25, opened 2016-09-23
Add errorreporting ruby integration tests

swcloud:


generated/ruby/test/test_errorreporting.rb
```diff
@@ -0,0 +1,101 @@
...
+PROJECT_NAME = "gapic-test".freeze
+SERVICE = "default".freeze
+VERSION = "20160920t135819".freeze
+USER_NAME = "testuser".freeze
+#
+# Integration test for ReportErrorsServiceApi.report_error_event
+#
+# Report first two error events (to be one group) and third event (to be another group)
+require "google/cloud/error_reporting/v1beta1/report_errors_service_api"
+require "google/devtools/clouderrorreporting/v1beta1/report_errors_service_pb"
```
jmuk: I am not so sure about the style of how this file is formatted. Is that because some section could be extracted through a tool or something?  If nothing is wrong, I'd recommend to:
```ruby
# Copyright header...

require '...'

PROJECT_NAME = ....

# ... actual code
```

generated/ruby/test/test_errorreporting.rb
```diff
@@ -0,0 +1,101 @@
...
+ReportErrorsServiceApi = Google::Cloud::ErrorReporting::V1beta1::ReportErrorsServiceApi
+ReportedErrorEvent = Google::Devtools::Clouderrorreporting::V1beta1::ReportedErrorEvent
+ServiceContext = Google::Devtools::Clouderrorreporting::V1beta1::ServiceContext
+ErrorContext = Google::Devtools::Clouderrorreporting::V1beta1::ErrorContext
+HttpRequestContext = Google::Devtools::Clouderrorreporting::V1beta1::HttpRequestContext
+SourceLocation = Google::Devtools::Clouderrorreporting::V1beta1::SourceLocation
```
jmuk: I'm not sure how much those lines help the code. Could it be
```ruby
V1beta1 = Google::Devtools::Clouderrorreporting::V1beta1
```

and those classes can be referred like `V1beta1::ServiceContext`, and I think that's not that bad.

generated/ruby/test/test_errorreporting.rb
```diff
@@ -0,0 +1,101 @@
...
+
+report_errors_service_api = ReportErrorsServiceApi.new
+formatted_project_name = ReportErrorsServiceApi.project_path(PROJECT_NAME)
+http_request_context = HttpRequestContext.new(method: "GET", url: "https://20160920t135819-dot-gapic-test.appspot.com")
+source_location1 = SourceLocation.new(file_path: "foo/bar", line_number: 20, function_name: "doGet")
+source_location2 = SourceLocation.new(file_path: "foo/bar", line_number: 30, function_name: "doSet")
+p "Reporting error event 1"
```
jmuk: `p` isn't a great tool for this purpose -- use `puts` instead

generated/ruby/test/test_errorreporting.rb
```diff
@@ -0,0 +1,101 @@
...
+#
+require "google/cloud/error_reporting/v1beta1/error_group_service_api"
+require "google/devtools/clouderrorreporting/v1beta1/error_group_service_pb"
+ErrorGroupServiceApi = Google::Cloud::ErrorReporting::V1beta1::ErrorGroupServiceApi
+ErrorGroup = Google::Devtools::Clouderrorreporting::V1beta1::ErrorGroup
+TrackingIssue = Google::Devtools::Clouderrorreporting::V1beta1::TrackingIssue
+error_group_service_api = ErrorGroupServiceApi.new
+formatted_group_name = ErrorGroupServiceApi.group_path(PROJECT_NAME, error_group_id)
+
+p "Getting error group " + formatted_group_name
```
jmuk: Would be nicer to use string embedding, like `"Getting error group #{formatted_group_name}"`

generated/ruby/test/test_errorreporting.rb
```diff
@@ -0,0 +1,101 @@
...
+response = error_group_service_api.get_group(formatted_group_name)
+p [response.name, response.group_id].join(" ")
+p "Associated issues"
+response.tracking_issues.each {|issue| p issue.url}
+
+p "Updating error group " + formatted_group_name
+issues = [TrackingIssue.new(url: "https://github.com/testuser/project/issues/update1")]
+error_group = ErrorGroup.new(name: formatted_group_name, group_id: error_group_id, tracking_issues: issues)
+response = error_group_service_api.update_group(error_group)
+p "Update failed" if response != error_group
```
jmuk: `unless response == error_group`

generated/ruby/test/test_errorreporting.rb
```diff
@@ -0,0 +1,101 @@
...
+ErrorStatsServiceApi = Google::Cloud::ErrorReporting::V1beta1::ErrorStatsServiceApi
+QueryTimeRange = Google::Devtools::Clouderrorreporting::V1beta1::QueryTimeRange
+
+p "Listing all the error groups"
+error_stats_service_api = ErrorStatsServiceApi.new
+formatted_project_name = ErrorStatsServiceApi.project_path(PROJECT_NAME)
+time_range = QueryTimeRange.new(period: :PERIOD_1_DAY)
+error_group_id = 0
+error_stats_service_api.list_group_stats(formatted_project_name, time_range).each_with_index do |error_group_stat, index|
+  p [error_group_stat.group.name, error_group_stat.group.group_id, error_group_stat.count].join(" ")
```
jmuk: I prefer to use a format instead of creating a style itself. It could be `puts error_group_stat.to_yaml` (format in YAML) or `puts JSON.pretty_generate(error_group_stat)` Same thing can be applied all among the file.

generated/ruby/test/test_errorreporting.rb
```diff
@@ -0,0 +1,101 @@
...
+require "google/devtools/clouderrorreporting/v1beta1/error_stats_service_pb"
+ErrorStatsServiceApi = Google::Cloud::ErrorReporting::V1beta1::ErrorStatsServiceApi
+QueryTimeRange = Google::Devtools::Clouderrorreporting::V1beta1::QueryTimeRange
+
+p "Listing all the error groups"
+error_stats_service_api = ErrorStatsServiceApi.new
+formatted_project_name = ErrorStatsServiceApi.project_path(PROJECT_NAME)
+time_range = QueryTimeRange.new(period: :PERIOD_1_DAY)
+error_group_id = 0
+error_stats_service_api.list_group_stats(formatted_project_name, time_range).each_with_index do |error_group_stat, index|
```
jmuk: Nicer to avoid exceeding 80 characters in a line.

generated/ruby/test/test_errorreporting.rb
```diff
@@ -0,0 +1,101 @@
+#
```
jmuk: The file name itself -- python seems to have `generated/python/test/vision.py`, why does this have `test_` prefix?

generated/ruby/test/test_errorreporting.rb
swcloud: Do we need to add v1beta1.rb in google/devtools/clouderrorreporting as well?

jmuk: That is independent from this. My suggestion was to create a shortened name for the parent module instead of the individual classes.

generated/ruby/test/errorreporting.rb
```diff
@@ -0,0 +1,160 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Integration tests for vkit generated client library for
+# Stackdriver ErrorReporting APIs.
+#
+require "google/cloud/error_reporting/v1beta1"
+
+PROJECT_NAME = "gapic-test".freeze
```
jmuk: Is it okay to embed this project name as a constant? Can we get the name from the environment variable (i.e. `ENV['PROJECT_ID']`?)

generated/ruby/test/errorreporting.rb
```diff
@@ -0,0 +1,160 @@
...
+SERVICE = "default".freeze
+VERSION = "20160920t135819".freeze
```
jmuk: just curious: what does this version string mean?

generated/ruby/test/errorreporting.rb
```diff
@@ -0,0 +1,160 @@
...
+ServiceContext = Grpc::ServiceContext
+ErrorContext = Grpc::ErrorContext
+HttpRequestContext = Grpc::HttpRequestContext
+SourceLocation = Grpc::SourceLocation
+
+
+formatted_project_name = ReportErrorsServiceApi.project_path(PROJECT_NAME)
+http_request_context = HttpRequestContext.new(
+  method: "GET",
+  url: "https://20160920t135819-dot-gapic-test.appspot.com")
```
jmuk: Does this URL have to be an appspot.com URL? Since it's a test, some "example.com" URL might make sense if that's okay.

generated/ruby/test/errorreporting.rb
swcloud: Version is a label indicating a specific running instance of for example GAE.

generated/ruby/test/errorreporting.rb
swcloud: It can be any URL in this test code.

generated/ruby/test/errorreporting.rb
swcloud: It looks like we want to write the test in a more formal way. I thought this test was just some simple code to verify the API calls via client libraries to just work, in other words, sort of smoke test.  I think formal tests should be done via a test framework with automation capability. So I'm not sure we should invest too much time on this code.

generated/ruby/google-cloud-ruby/google-cloud-error_reporting/lib/google/cloud/error_reporting/v1beta1.rb
```diff
@@ -0,0 +1,19 @@
+
```
jmuk: This file also needs to be in the other PR.

swcloud: I agree, but the reason why I kept it here because in order to run the modified integration test (with new way to call 'require'), I have to have the v1beta1.rb here.

jmuk: I'm not sure I got your point. That does not mean this file has to be on this PR. Of course you will have to customize the local environment when you are trying to do that before everything is merged, but that does not justify the PR structures.

swcloud: ok. I moved it to https://github.com/googleapis/api-client-staging/pull/38

generated/ruby/test/errorreporting.rb
```diff
@@ -0,0 +1,165 @@
+# Copyright 2016 Google Inc. All rights reserved.
```
landrito: File location is a bit misleading. The test is not generated but hand written. Might want to consider moving the file to tests/ruby/google-cloud-ruby/google-cloud-error_reporting/

generated/ruby/test/errorreporting.rb
```diff
@@ -0,0 +1,165 @@
...
+  tracking_issues: issues)
+response = error_group_service_api.update_group(error_group)
+puts "Update failed" unless response == error_group
+
+#
+# Integration test for ErrorStatsServiceApi.delete_events
+#
+puts "Deleting all the events in project " + PROJECT_NAME
+formatted_project_name = ErrorStatsServiceApi.project_path(PROJECT_NAME)
+error_stats_service_api.delete_events(formatted_project_name)
```
landrito: Wrap this in an ensure in case any of the above throws an exception.

generated/ruby/test/errorreporting.rb
```diff
@@ -0,0 +1,165 @@
...
+PROJECT_NAME = ARGV[0].freeze
+SERVICE = ARGV[1].freeze
+VERSION = ARGV[2].freeze
+USER_NAME = "testuser".freeze
+
+#
+# Integration test for ReportErrorsServiceApi.report_error_event
+#
+# Report first two error events (to be one group) and third event (to be another group)
+ReportErrorsServiceApi =
```
landrito: Wrap in a begin statement so that the delete events can be ensured.

generated/ruby/test/errorreporting.rb
swcloud: I agree. This was following https://github.com/googleapis/api-client-staging/blob/master/generated/python/test/vision.py.

But I do think the location should be in a different place.

landrito:
cc @bjwatson. What are your thoughts on moving the tests to a top level test directory? Personally, I feel that file location inside generated is a bit misleading. The test is not generated, but hand written.

swcloud:
I think Brian already agreed to move the hand-written test to the top-level 'test' directory in a separate email thread.

PR #27, opened 2016-09-23
DO NOT MERGE: Add streaming to Speech.

shinfan:
- This is an example PR that shows bi-directional streaming support in vkit java. 
   - If the streaming is one-directional (the request is not StreamObserver), parameter flattening will be applied as the same as other API calls.
- Right now we just expose the grpc StreamObserver directly.


generated/java/google-cloud-speech-v1beta1/src/main/java/com/google/cloud/speech/spi/v1beta1/SpeechApi.java
```diff
@@ -146,6 +154,11 @@ public void close() throws IOException {
     }
   }
 
+  public final StreamObserver<StreamingRecognizeResponse> streamingRecognize(
+      StreamObserver<StreamingRecognizeRequest> streamObserver) {
+    return streamingRecognizeCallable.call(streamObserver);
+  }
```
garrettjonesgoogle: I'd like to see the callable variant too.

generated/java/google-cloud-speech-v1beta1/src/main/java/com/google/cloud/speech/spi/v1beta1/SpeechSettings.java
```diff
@@ -95,6 +100,8 @@
   private final SimpleCallSettings<SyncRecognizeRequest, SyncRecognizeResponse>
       syncRecognizeSettings;
   private final SimpleCallSettings<AsyncRecognizeRequest, Operation> asyncRecognizeSettings;
+  private final SimpleCallSettings<StreamObserver<StreamingRecognizeRequest>,
+      StreamObserver<StreamingRecognizeResponse>> streamingRecognizeSettings;
```
garrettjonesgoogle: Might we need a StreamingCallSettings?

generated/java/google-cloud-speech-v1beta1/src/main/java/com/google/cloud/speech/spi/v1beta1/SpeechApi.java
shinfan: Right now streamingRecognizeCallable is a SimpleApiCallable object. The logic is the same as the other simple API calls except the request/response types are StreamObserver<T>

shinfan:
Uploaded the smoke test to demonstrate the client interaction.

generated/java/google-cloud-speech-v1beta1/src/main/java/com/google/cloud/speech/spi/v1beta1/SpeechApi.java
shinfan: Added callable variant

generated/java/google-cloud-speech-v1beta1/src/main/java/com/google/cloud/speech/spi/v1beta1/SpeechApi.java
```diff
@@ -146,6 +154,16 @@ public void close() throws IOException {
...
 
+  public final StreamObserver<StreamingRecognizeRequest> streamingRecognize(
+      StreamObserver<StreamingRecognizeResponse> streamObserver) {
+    return streamingRecognizeCallable.call(streamObserver);
+  }
+
+  public final ApiCallable<StreamObserver<StreamingRecognizeResponse>,
+      StreamObserver<StreamingRecognizeRequest>> streamingRecognizeCallable() {
+    return streamingRecognizeCallable;
+  }
```
garrettjonesgoogle: I'm wondering if we should go fully minimalist with this and only provide the callable.

shinfan: I think the parameter/return types of streamingRecognize() is a bit of confusing so only providing the callable might be better. I have no objections to it.

shinfan:
cc @jmuk

PR #30, opened 2016-09-26
Add nodejs integration test for errorreporting API

swcloud:


generated/nodejs/test/errorreporting.js
```diff
@@ -0,0 +1,245 @@
...
+  };
+
+  console.log('Listing all the error groups');
+  var groups = [];
+  api.listGroupStats(formattedProjectName, timeRange).on('data',
+      function(errorGroupStat) {
+        console.log(errorGroupStat.group.name + ' ' +
+            errorGroupStat.group.group_id + ' ' +
+            errorGroupStat.count);
+        groups.push(errorGroupStat);
```
jmuk: It seems you don't have to have the full set of `groups`, right?
You can set the first state into a variable like `firstErrorGroup`, and invoke `end` to the stream so that the stream knows you don't need data anymore. That is:
```javascript
var stream = api.listGroupStats(formattedProjectName, timeRange);
stream.on('data', function(errorGroupStat) {
  console.log(...);
  // no more data.
  stream.end();
  listErrorGroupEvents(errorGroupStat.group.group_id);
});
```

swcloud: Well, I do want to list all the groups (two groups), but for listing the events in a group, I only want the first group, which contains two events.

generated/nodejs/test/errorreporting.js
```diff
@@ -0,0 +1,245 @@
...
+  });
+}
+
+createErrorEvents();
+setTimeout(function() {
+  listErrorGroupStats();
+}, 5000);
+setTimeout(function() {
+  deleteErrorEvents();
+}, 15000);
```
swcloud: Sounds good. I will use async.

generated/nodejs/test/errorreporting.js
```diff
@@ -44,7 +51,7 @@ var clouderrorreportingV1beta1 =
  * Report first two error events (to be one group) and third event
  * (to be another group)
  */
-function createErrorEvents() {
+function createErrorEvents(createErrorEventsDoneCallback) {
```
jmuk: nit: `done` is more idiomatic name

generated/nodejs/test/errorreporting.js
```diff
@@ -115,11 +105,36 @@ function createErrorEvents() {
...
     }
   };
-  api.reportErrorEvent(formattedProjectName, event3, function(err, response) {
+
+  // report the three events in parallel
+  async.eachOf([event1, event2, event3], function(event, index, callback) {
+    api.reportErrorEvent(formattedProjectName, event, function(err, response) {
+      if (err) {
+        console.error(err);
+        return;
```
jmuk: you should invoke `callback(err)`, otherwise `async` can't capture the errors correctly.

generated/nodejs/test/errorreporting.js
```diff
@@ -115,11 +105,36 @@ function createErrorEvents() {
...
+      console.log('Reporting error event %d' ,index + 1);
+      callback();
+    });
+  }, function(err) {
     if (err) {
       console.error(err);
       return;
     }
+    console.log('createErrorEventsDone');
+    createErrorEventsDoneCallback(null, 'createErrorEventsDone');
```
jmuk: I don't think the second argument is necessary.

generated/nodejs/test/errorreporting.js
```diff
@@ -115,11 +105,36 @@ function createErrorEvents() {
...
+      if (err) {
+        console.error(err);
+        return;
+      }
+      console.log('Reporting error event %d' ,index + 1);
+      callback();
+    });
+  }, function(err) {
     if (err) {
       console.error(err);
```
jmuk: Similarly, you will have to invoke `createErrorEventsDoneCallback` with the error.

generated/nodejs/test/errorreporting.js
```diff
@@ -115,11 +105,36 @@ function createErrorEvents() {
...
+}
+
+function listErrorInfo(listErrorInfoDoneCallback) {
+  async.waterfall([
+    listErrorGroupStats,
+    listErrorGroupEvents,
+    updateErrorGroup,
+    getErrorGroup
+  ], function(err, result) {
+    console.log('listErrorInfo' + result);
```
jmuk: `console.log` accepts multiple arguments. `console.log('listErrorInfo', result);` would look better.

generated/nodejs/test/errorreporting.js
```diff
@@ -146,11 +161,11 @@ function listErrorGroupStats() {
       }).on('end', function() {
         var firstErrorGroupId = groups[0].group.group_id;
```
jmuk: I still don't get why `groups` exist. Only the first element is used.

generated/nodejs/test/errorreporting.js
```diff
@@ -146,11 +161,11 @@ function listErrorGroupStats() {
...
         console.log('firstErrorGroupId = ' + firstErrorGroupId);
-        listErrorGroupEvents(firstErrorGroupId);
+        callback(null, firstErrorGroupId);
       });
```
jmuk: `.on('error', callback)` would be better to be added.

generated/nodejs/test/errorreporting.js
```diff
@@ -170,16 +185,39 @@ function listErrorGroupEvents(errorGroupId) {
...
+      url: "https://github.com/testuser/project/issues/update1"
+    }];
+  var errorGroup = {
+    name: formattedGroupName,
+    group_id: errorGroupId,
+    tracking_issues: issues
+  };
+  api.updateGroup(errorGroup, function(err, response) {
+    if (err) {
+      console.error(err);
```
jmuk: ditto: callback(err);

generated/nodejs/test/errorreporting.js
```diff
@@ -234,12 +250,10 @@ function deleteErrorEvents() {
       return;
     }
   });
+
+  console.log('deleteErrorEventsDone');
+  deleteErrorEventsCallback(null, 'deleteErrorEventsDone');
```
jmuk: Please move these lines inside of the callback.

generated/nodejs/test/errorreporting.js
swcloud: I created three events, the first two events belong to the first group, the third event belongs to the second group. For listEvents with a group, I chose the first group since there are two events in it.

jmuk: I am saying about `groups` variable declared at line 139.  `lsitGroupStats`'s `data` callback pushes new elements to `groups`, and then, only the first item is used in this line 147, and then, other elements are not used at all.

Is my understanding correct?  If so why are you pushing everything into an array?

swcloud: I guess I asked you about this. On ruby, for listGroupStats, since it returns an array, I can call eash_with_index and use index 0 to find out the first group's ide, but on node, with stream.on('data'), I don't know how to get the first group since I don't a way to use a index, so I just pushed all the groups into an array, so at stream.on('end'), I can use group[0] to get the first group's id.

test/nodejs/errorreporting/errorreporting.js
```diff
@@ -0,0 +1,255 @@
...
+      if (err) {
+        console.error(err);
+      }
+      callback(err);
+    });
+  }, function(err) {
+    if (err) {
+      console.error(err);
+    }
+    console.log('createErrorEvents Done');
```
jmuk: `else { console.log('createErrorEvents Done'); }` ? Otherwise, it will be printed out when errors (did you intend that?)

generated/nodejs/test/errorreporting.js
jmuk:
```javascript
var firstGroup;
api.listGroupStats(...).on('data', function(errorGroupStat) {
   console.log(...);
   if (firstGroup === undefined) {
    firstGroup = errorGroupStat;
   }
}).on('end', function() {
  if (firstGroup === undefined) {
     // log error -- no groups
  } else {
    ...
  }
})...
```

test/nodejs/errorreporting/errorreporting.js
```diff
@@ -0,0 +1,255 @@
...
+      {timeRange: timeRange}).on('data', function(errorEvent) {
+        console.log(errorEvent.message + ' ' +
+            errorEvent.service_context.service + ' ' +
+            errorEvent.service_context.version);
+        console.log(errorEvent.context.user);
+        console.log(errorEvent.context.http_request.method + ' ' +
+            errorEvent.context.http_request.url);
+        console.log(errorEvent.context.report_location.file_path + ' ' +
+            errorEvent.context.report_location.line_number + ' ' +
+            errorEvent.context.report_location.function_name);
```
jmuk: It is still a bit unclear how beneficial you format the result.

Indeed Node's default `console.log` does some level of pretty-printing, but it does not print greatly on some deeper elements, but isn't it sufficient to use a pretty-printing JSON format?
```javascript
console.log(JSON.stringify(errorEvent, null, 4));
```
might be good enough (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify)

test/nodejs/errorreporting/errorreporting.js
```diff
@@ -0,0 +1,255 @@
...
+/**
+ * Integration test for ErrorGroupServiceApi.getGroup,
+ *                     ErrorGroupServiceApi.updateGroup
+ * @param {String} errorGroupId - error group id
+ */
+function getErrorGroup(errorGroupId, callback) {
+  var api = clouderrorreportingV1beta1.errorGroupServiceApi();
+  var formattedGroupName = api.groupPath(PROJECT_NAME, errorGroupId);
+
+  console.log('Getting error group ' + formattedGroupName);
```
jmuk: `console.log('Getting error group', formattedGroupName);`

test/nodejs/errorreporting/errorreporting.js
```diff
@@ -0,0 +1,255 @@
...
+  api.getGroup(formattedGroupName, function(err, response) {
+    if (err) {
+      console.error(err);
+      callback(err);
+      return;
+    }
+    console.log(response.name + ' ' + response.group_id);
```
jmuk: `console.log(response.name, response.group_id);`

test/nodejs/errorreporting/errorreporting.js
```diff
@@ -0,0 +1,255 @@
...
+    console.log('Associated issues');
+    for (var i = 0; i < response.tracking_issues.length; i++) {
+      console.log(response.tracking_issues[i].url);
+    }
+    callback(null, 'Done');
```
jmuk: Why `'Done'`?

This seems to be `result` in `listErrorInfo` at line 134, but that's not very informative. Returning with `response`?  Or `console.log` in line 134 might not be necessary.

test/nodejs/errorreporting/errorreporting.js
swcloud: Yes. That's my intention. I want to print 'createErrorEvents Done' regardless

test/nodejs/errorreporting/errorreporting.js
swcloud: The purpose for adding 'Done' as result is trying to be consistent with createErrorEvents and deleteErrorEvents in terms of printing the major steps with a 'Done'.

jmuk: In that case `console.log('listErrorInfo Done');` in line 134 and removing this would be much clearer (also, it will print out `'listErrorInfo null` in case of error which I think unintentional).

test/nodejs/errorreporting/errorreporting.js
```diff
@@ -0,0 +1,255 @@
...
+  api.deleteEvents(formattedProjectName, function(err, response) {
+    if (err) {
+      console.error(err);
+    }
+    console.log('deleteErrorEvents Done');
+    done(err);
+  });
+}
+
+async.series([createErrorEvents, listErrorInfo, deleteErrorEvents]);
```
jmuk: Print out the error at last if something got wrong in the middle?
```javascript
async.series([createErrorEvents, listErrorInfo, deleteErrorEvents], function(err) {
  if (err) {
    console.log(err);
  }
});
```

test/nodejs/errorreporting/errorreporting.js
swcloud: Done.

test/nodejs/errorreporting/errorreporting.js
swcloud: Done.

PR #40, opened 2016-09-28
Updates to vision.

Landrito:


jmuk:
Please also create src/v1/index.js -- either through packman or through hand-written.

PR #50, opened 2016-09-29
Auto-generate Speech for Python

bjwatson:
Include GAPIC LRO layer, and updates to googleapis-common-protos.


bjwatson:
@daspecster This contains the Speech and LRO updates you need. I'm just waiting on a review from @geigerj (along with everything else I've thrown at him).

PR #55, opened 2016-09-30
Auto-generate Python Bigtable and Datastore

bjwatson:


bjwatson:
@tseaver @dhermes I'll publish this once I get approval from @geigerj (although I have been throwing a lot of these at him today).

dhermes:
Nice!

bjwatson:
@geigerj FYI, the common-protos dependency updates I slipped in here for other packages is what I actually published them with.

geigerj:
@bjwatson FYI: there are two more Bigtable services (admin): https://github.com/googleapis/googleapis/tree/master/google/bigtable/admin/v2. Presumably we will also want to wrap them before we call Bigtable "done".

PR #60, opened 2016-09-30
Longrunning updates

bjwatson:
Fix directory name to match package name, and bump version number due to a boneheaded mistake I made.


geigerj:
What was the change? I see that the directory in `api-client-staging` was wrong (missed that in my last review), but has the package itself changed at all?

bjwatson:
Just a directory name change. No package change apart from the version bump.

geigerj:
Ok -- I don't think that needs a version bump, then. But LGTM.

bjwatson:
The version bump was because I burned `0.11.1` due to a mistake in publishing to PyPI.

PR #62, opened 2016-09-30
Go: Add files from cloud.google.com/go repo

### LINKED ISSUE #33: Generate Go clients
Generate the active Google APIs for Go.

pongad:
These files are already committed into Go cloud repo.
After this gets submitted, changes reflecting new features in toolkit
can be added in separate PRs to ease reviewing.

Fixes #33.


garrettjonesgoogle:
For several languages at least, we have been using the api_name value from the artman config for the subdirectory. For your case, for the example of pubsub, this would look like:

`generated/go/google-pubsub-v1/cloud.google.com/...`

I think I want to push all of the languages to use that pattern.

pongad:
@garrettjonesgoogle I seem to have dun goof. PTAL

garrettjonesgoogle:
LGTM

PR #64, opened 2016-09-30
Adding java/google-longrunning

garrettjonesgoogle:


garrettjonesgoogle:
PTAL again - I removed 'cloud' from the package name.

PR #71, opened 2016-10-02
Include test sample files, so they are repeatable by others

bjwatson:
This issue was raised, because @daspecster was trying to reproduce our successful test results.


daspecster:
Thanks @bjwatson!

For Vision, handling the response for a more complex request like Face Detection might be good to demo here.

For Speech, there are 3 types of requests that are all handled slightly differently. `AsyncRecognize`, `SyncRecognize` and `StreamingRecognize`. The setup for each of those is slightly different and the handling of the responses as well.

I'm not sure how important having these examples would be for new users?
The tests that you guys are working on would also be a great place to see how these are exercised.

bjwatson:
@daspecster We aren't planning to do any more work on Vision and Speech here. These smoke tests are temporary, until your google-cloud-python system tests cover us and/or we implement auto-generated testing (which we already have for Java).

bjwatson:
@geigerj I updated this PR as you requested offline. PTAL

PR #77, opened 2016-10-05
Add NodeJS Trace integration test

geigerj:

PR #78, opened 2016-10-05
Add list method to Python Trace integration test

geigerj:
This makes the test correspond to the NodeJS test in
https://github.com/googleapis/api-client-staging/pull/77

### DIFF
--- a/generated/python/test/trace.py
+++ b/generated/python/test/trace.py
@@ -28,10 +28,19 @@ span = trace_pb2.TraceSpan(span_id=span_id,kind=span_kind,name=SPAN_NAME,
 trace = trace_pb2.Trace(project_id=PROJECT_ID, trace_id=id, spans=[span])
 traces = trace_pb2.Traces(traces=[trace])
 
+print 'testing patch'
 api.patch_traces(PROJECT_ID, traces)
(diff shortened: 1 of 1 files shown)


generated/python/test/trace.py
```diff
@@ -28,10 +28,19 @@
 trace = trace_pb2.Trace(project_id=PROJECT_ID, trace_id=id, spans=[span])
 traces = trace_pb2.Traces(traces=[trace])
 
+print 'testing patch'
 api.patch_traces(PROJECT_ID, traces)
+print 'patched'
 
+# wait for patch to propagate
```
jmuk: As I commented in #77, I'm even happier if you leave a link to a document to justify the sleep (and why 1 second is sufficient), if such document exists.

jmuk: As I responded in #78 -- this new comment is just good if there is no docs.

bjwatson: I'm okay with a comment that says "seems to be long enough to avoid races". Remember that these are ad hoc smoke tests that will become obsolete this quarter as we do the automated unit and smoke testing.

bjwatson:
LGTM

PR #95, opened 2016-11-07
Regenerate speech_api.js with gRPC-streaming support.

jmuk:


bjwatson:
Looks okay, but @landrito would better understand what's being done here.

generated/nodejs/google-cloud-node/packages/speech/src/v1beta1/speech_api.js
```diff
@@ -208,6 +211,35 @@ SpeechApi.prototype.asyncRecognize = function(request, options, callback) {
...
+ *
+ * @example
+ *
+ * var api = speechV1beta1.speechApi();
+ * var stream = api.streamingRecognize().on('data', function(response) {
+ *     // doThingsWith(response);
+ * });
+ * var request = {};
+ * // Write request objects.
+ * stream.write(request);
```
jmuk: The handling is written in line 230, `// doThingsWith(response);`, which is common among other APIs.
You prefer that after writing requests?

bjwatson: Oh, right. I've been doing imperative languages for too long. :)

How about producing streaming requests?

landrito: @bjwatson
This section is what makes the api requests.
```js
var request = ...;
stream.write(request);
```
Is this what you mean about producing requests?

landrito: I see. The example should be sufficient for consumption.
```js
stream.on('data', function(response) {
  // doThingsWith(response);
});
```
For using an input stream, are you suggesting we add something like such:
```js
var requestStream = ...;
requestStream.on('data', function(item) {
   var request = {name: item.name ... };
   stream.write(request);
});
```

PR #103, opened 2016-11-17
Datastore v0.13.1

bjwatson:
Rebuilt with latest `toolkit` and `grpcio` code.

FYI @dhermes. I'll publish this once @geigerj approves. Note that I never published the previous 0.11.1 version, since I wanted to confirm the package name change with the GAPIC team, so that we can be more consistent with how `google-cloud-python` names packages.


generated/python/gapic-google-cloud-datastore-v1/setup.py
```diff
@@ -9,15 +9,15 @@
 import sys
 
 install_requires = [
-    'googleapis-common-protos>=1.3.5, <2.0.0dev',
-    'google-gax>=0.14.1, <0.15.0dev',
-    'grpc-google-datastore-v1>=0.11.1, <0.12.0dev',
-    'oauth2client>=2.0.0, <4.0.0dev',
+    'googleapis-common-protos>=1.5.0, <2.0dev',
+    'google-gax>=0.14.1, <0.15dev',
```
dhermes: @bjwatson If `google-gax==0.15` just went out today, why doesn't this version depend on it?

bjwatson: @dhermes Jacob did the necessary codegen changes today to make that work. See https://github.com/googleapis/api-client-staging/pull/106

PR #106, opened 2016-11-18
Regenerate Datastore and Language gRPC and GAPIC 0.14.x packages

bjwatson:
FYI @dhermes. This should allow you try using [`google-auth`](https://pypi.python.org/pypi/google-auth), rather than [`oauth2client`](https://pypi.python.org/pypi/oauth2client).


bjwatson:
@geigerj I made another change based on a [comment from @dhermes](https://github.com/GoogleCloudPlatform/google-cloud-python/issues/2746#issuecomment-261597513). PTAL

geigerj:
@bjwatson What is the status of this?
* Is the most recent handedit this a beta blocker?
* Do we need to automate it?

If possible, I'd prefer not to put any more hand edits into our code and instead put the pressure on gRPC to release 1.1, which should result in this issue being resolved upstream.

bjwatson:
@geigerj Thanks for the reminder. If we don't do this now, then @dhermes can't update the hand-written Datastore code to use GAPIC for gRPC. 

@nathanielmanistaatgoogle @soltanmm-google How soon will gRPC 1.1 be ready for release? We're particularly interested in https://github.com/grpc/grpc/pull/8246.

geigerj:
@bjwatson Assuming this is not a beta blocker (since we don't want to do plumbing switches right before beta), I'm pushing for this just to happen with gRPC 1.1 instead of adding more hand edits to the generated proto/gRPC code.

bjwatson:
@geigerj Agreed.

soltanmm-google:
grpc/grpc#8246 is going to be reverted and/or totally changed up due to grpc/grpc#8753.

geigerj:
@soltanmm-google Note that the alternative originally proposed in https://github.com/grpc/grpc/issues/8056, i.e., include all gRPC code in a `try/except` block, would work for us as well.

bjwatson:
@soltanmm-google The approach that @geigerj

soltanmm-google:
@geigerj Ah, okay, I think that part hasn't changed.

bjwatson:
@geigerj I re-generated this using `grpcio 1.0.2rc0`, which avoids the need for hand-edits with the gRPC code. PTAL

bjwatson:
@dhermes Can I get a quick check from you on Datastore and Language, before I merge and publish?

dhermes:
Sure

dhermes:
Looks fine, not sure what to look for in the diffs. I was curious that, for example `language_service_pb2_grpc.py` never is imported anywhere (or maybe it is, in a suppressed diff).

geigerj:
@dhermes @bjwatson IIUC the gRPC methods in `language_service_pb2` are deprecated. We will switch to importing `language_service_pb2_grpc.py` instead to access these methods in the GAPIC layer in the near future.

bjwatson:
@geigerj is correct. Those files were introduced in https://github.com/grpc/grpc/issues/8056, but aren't used just yet.

bjwatson:
@dhermes I have published these packages. The Datastore links can be found in https://github.com/googleapis/googleapis/pull/220

PR #109, opened 2016-11-22
Java: Updating LongRunning

garrettjonesgoogle:


generated/java/google-longrunning-v1/src/main/java/com/google/longrunning/OperationsClient.java
```diff
@@ -198,16 +201,17 @@ public final OperationsSettings getSettings() {
...
+   *   String formattedName = OperationsClient.formatOperationPathName("[OPERATION_PATH]");
+   *   Operation response = operationsClient.getOperation(formattedName);
    * }
    * </code></pre>
    *
    * @param name The name of the operation resource.
    * @throws com.google.api.gax.grpc.ApiException if the remote call fails
    */
   public final Operation getOperation(String name) {
+    OPERATION_PATH_PATH_TEMPLATE.validate(name, "getOperation");
```
michaelbausor: Is it the case that the operation name must always be formatted this way? The comment in the proto seems to hedge a bit: https://github.com/googleapis/googleapis/blob/master/google/longrunning/operations.proto#L84

garrettjonesgoogle: Good catch - I think we should remove the collection configuration for this - https://github.com/googleapis/googleapis/blob/master/google/longrunning/longrunning_gapic.yaml#L24 - I'll send a PR for it.

garrettjonesgoogle: https://github.com/googleapis/googleapis/pull/201

garrettjonesgoogle:
Updated, PTAL

michaelbausor:
LGTM

PR #128, opened 2016-12-07
Regenerate PHP

michaelbausor:
- Get GAX version from GAX instead of hard coded
- Add note about experimental status
- Proto comment updates


generated/php/google-cloud-error-reporting-v1beta1/src/Errorreporting/V1beta1/ErrorGroupServiceClient.php
```diff
@@ -18,6 +18,10 @@
  * This file was generated from the file
  * https://github.com/google/googleapis/blob/master/google/devtools/clouderrorreporting/v1beta1/error_group_service.proto
  * and updates to that file get reflected here through a refresh process.
+ *
+ * EXPERIMENTAL: the code generation tool used to create this class has not yet been declared
```
garrettjonesgoogle: Sorry I didn't notice this on the toolkit CL, but the release level of the client libraries isn't really related to the release level of the code generation tool. I think it should just say that this client library isn't beta-level yet.

michaelbausor:
PTAL

garrettjonesgoogle:
LGTM

PR #130, opened 2016-12-07
Mention documentation in Python logging README

geigerj:


generated/python/gapic-google-cloud-logging-v2/setup.py
```diff
@@ -17,7 +17,7 @@
 
 setup(
     name='gapic-google-cloud-logging-v2',
-    version='0.14.0',
+    version='0.14.1',
```
ethanbao: Why do you include this change in this PR?

geigerj: I need to republish the package for this README to show up on PyPI. I cannot republish version 0.14.0; it is already published.

ethanbao:
LGTM

geigerj:
@bjwatson Could you take a look?

PR #131, opened 2016-12-08
Updates to the speech client using new LRO wrapper.

landrito:
cc @swcloud

### DIFF
--- a/generated/ruby/google-cloud-ruby/google-cloud-speech/lib/google/cloud/speech/v1beta1/speech_client.rb
+++ b/generated/ruby/google-cloud-ruby/google-cloud-speech/lib/google/cloud/speech/v1beta1/speech_client.rb
@@ -26,6 +26,9 @@ require "json"
 require "pathname"
 
 require "google/gax"
+require "google/gax/operation"
+require "google/longrunning/operations_api"
+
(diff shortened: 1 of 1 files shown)


thagomizer:
LGTM.

hxiong388:
Looks good.

Nit: In the code samples, the `op.results` is really an V1beta1:AsyncRecognizeResponse object. The actual results is in `op.results.results`, which may be confusing to users. We should probably fix that in gax-ruby.

landrito:
Would `op.response` be a better alternative? This has the benefit of mirroring the [operations proto](https://github.com/googleapis/googleapis/blob/6ae2d424245deeb34cf73c4f7aba31f1079bcc40/google/longrunning/operations.proto#L114), but has the detriment of being confused with the api call response.

hxiong388:
@landrito That's what I thought at first, but then I realize the LRO implementation is being correctly transparent of the request/response body. It just happens in this case the operation result is a wrapped gRPC struct with an addition "results". I think it's enough to just the clarify that in the example code,  like `op_results = op.results`. 

This is just a small thing, I think it's still ok even if you leave it be.

landrito:
Gotcha. I'll make the small change in toolkit!

PR #139, opened 2016-12-13
Regenerate java

michaelbausor:
Bump grpc version to 0.1.5
Include latest toolkit ( and https://github.com/googleapis/toolkit/pull/844)

### DIFF
--- a/generated/java/google-cloud-error-reporting-v1beta1/src/test/java/com/google/cloud/errorreporting/spi/v1beta1/ErrorGroupServiceTest.java
+++ b/generated/java/google-cloud-error-reporting-v1beta1/src/test/java/com/google/cloud/errorreporting/spi/v1beta1/ErrorGroupServiceTest.java
@@ -79,10 +79,9 @@ public class ErrorGroupServiceTest {
   @Test
   @SuppressWarnings("all")
   public void getGroupTest() {
-    GroupName name = GroupName.create("[PROJECT]", "[GROUP]");
+    String name = "name3373707";
     String groupId = "groupId506361563";
-    ErrorGroup expectedResponse =
-        ErrorGroup.newBuilder().setNameWithGroupName(name).setGroupId(groupId).build();
+    ErrorGroup expectedResponse = ErrorGroup.newBuilder().setName(name).setGroupId(groupId).build();
     mockErrorGroupService.addResponse(expectedResponse);
 
     GroupName groupName = GroupName.create("[PROJECT]", "[GROUP]");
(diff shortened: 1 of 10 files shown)


generated/java/google-cloud-error-reporting-v1beta1/src/test/java/com/google/cloud/errorreporting/spi/v1beta1/ErrorGroupServiceTest.java
```diff
@@ -79,10 +79,9 @@ public void tearDown() throws Exception {
   @Test
   @SuppressWarnings("all")
   public void getGroupTest() {
-    GroupName name = GroupName.create("[PROJECT]", "[GROUP]");
+    String name = "name3373707";
     String groupId = "groupId506361563";
-    ErrorGroup expectedResponse =
-        ErrorGroup.newBuilder().setNameWithGroupName(name).setGroupId(groupId).build();
+    ErrorGroup expectedResponse = ErrorGroup.newBuilder().setName(name).setGroupId(groupId).build();
```
garrettjonesgoogle: What is the reason for this change?

michaelbausor: Previously, the resource names to be used here were determined by constructing a collection of field configs for fields on the method response object. However this was insufficient, because for some methods (LRO, page streaming) the method response object is not what we need. The general solution is instead to create field configs for all fields and messages with resource names. The toolkit PR (https://github.com/googleapis/toolkit/pull/844) does this in ApiConfig.

However, there is a problem with the logic (https://github.com/googleapis/toolkit/pull/844/files#diff-a0f27b8e4c41d2ebf887482b72314a58R57) for finding all messages defined in proto files - it seems to miss out messages such as ErrorGroup which is defined in a proto file that does not contain an interface.

This logic is borrowed from what C# uses to generate their resource name types, so I suspect that C# will have problem augmenting the ErrorGroup type and similar proto types with getters and setters, but I haven't confirmed this.

So basically, this is a problem that needs to be fixed, and I need to confirm that C# is affected and involve @chrisdunelm in the solution. I can work on fixing this either before or after this PR.

garrettjonesgoogle: If this refresh isn't urgent then let's fix it in this PR.

michaelbausor:
PTAL

garrettjonesgoogle:
LGTM

garrettjonesgoogle:
@michaelbausor , there are no generated grpc maven packages for most APIs at 0.1.5, so this doesn't build for me.

michaelbausor:
I went through the release process but it seems like I missed the last step, they are now released

PR #143, opened 2016-12-21
Regenerate PHP

michaelbausor:
- Updates grpc stub name
- Updates docs with backslashes
- Adds support for passing in stub creation function


garrettjonesgoogle:
LGTM

PR #144, opened 2016-12-21
Generate Logging external service types

geigerj:
Updates https://github.com/GoogleCloudPlatform/google-cloud-python/issues/2572
Updates https://github.com/googleapis/googleapis/issues/187


generated/python/grpc-google-service-logging-types-v1/README.rst
```diff
@@ -0,0 +1,5 @@
+gRPC library for Stackdriver Logging External Types
```
bjwatson: I love having real API names in the docs now!

generated/python/grpc-google-service-logging-types-v1/setup.py
```diff
@@ -0,0 +1,43 @@
...
+https://packaging.python.org/en/latest/distributing.html
+https://github.com/pypa/sampleproject
+"""
+
+import setuptools
+
+from setuptools import setup, find_packages
+
+install_requires = [
+  'oauth2client>=2.0.0, <4.0dev',
```
bjwatson: We should start setting the upper-bound at 5.0dev, but that can be done across-the-board in a separate PR.

generated/python/grpc-google-service-logging-types-v1/setup.py
```diff
@@ -0,0 +1,43 @@
...
+]
+
+setuptools.setup(
+  name='grpc-google-cloud-service-logging-types-v1',
+  version='0.14.0',
+  author='Google Inc',
+  author_email='googleapis-packages@google.com',
+  classifiers=[
+    'Intended Audience :: Developers',
+    'Development Status :: 3 - Alpha',
```
bjwatson: Alpha is fine for now since we might get breaking feedback, but we probably want to declare Beta on this before long given that Logging itself is already in Beta.

generated/python/grpc-google-service-logging-types-v1/setup.py
```diff
@@ -0,0 +1,43 @@
...
+    'Programming Language :: Python :: 3.4',
+    'Programming Language :: Python :: 3.5',
+    'Programming Language :: Python :: Implementation :: CPython',
+  ],
+  description='GRPC library for the Stackdriver Logging External Types service',
+  long_description=open('README.rst').read(),
+  install_requires=install_requires,
+  license='Apache-2.0',
+  packages=find_packages(),
+  namespace_packages=['google.cloud.grpc.appengine.legacy', 'google.cloud.grpc.appengine.logging', 'google.cloud.grpc.appengine', 'google.cloud.grpc', 'google.cloud', 'google'],
```
bjwatson: Should `'google.cloud.grpc.appengine.logging.v1'` be in this list?

bjwatson:
@lukesneeringer This is an example of our Python gRPC generated package (the lowest level in the client library stack). This contains extra types supported by our Logging service, which we just released to Beta a couple weeks ago.

generated/python/grpc-google-service-logging-types-v1/setup.py
geigerj: We don't currently declare version-number packages as namespace packages; this logic originated in [Packman](https://github.com/googleapis/packman/blob/378f7c479df0e0f7d243d06d48c4a60db5c47621/lib/packager.js#L223), and I later ported the same behavior to toolkit. I can't currently think of any reason to keep this behavior, though -- I'll file an issue in toolkit and we can discuss there whether this should be changed.

bjwatson: I saw the issue. Thanks!

PR #148, opened 2017-01-04
Regenerate PHP with unit tests

michaelbausor:
- Add composer file to allow tests to be run
- Regenerate all APIs, with unit tests added

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,7 @@ generated/java/*/build
 
 # Node
 node_modules/
+
+# PHP
+vendor/
+composer.lock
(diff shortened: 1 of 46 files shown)


shinfan:
Is it necessary to auto-gen the composer file for PHP? In NodeJS and Java we auto-gen the build files.

michaelbausor:
Good point. We don't need it for the staging repo, as it looks like composer doesn't support inheritance of composer.json files, so we will still need the manually created composer.json in the root. But we do want it so that an individually generated API is testable on its own, which requires a composer.json file

garrettjonesgoogle:
LGTM

PR #149, opened 2017-01-06
Exclude pyc files from all GAPIC packages

geigerj:
Updates https://github.com/GoogleCloudPlatform/google-cloud-python/issues/2913, https://github.com/googleapis/toolkit/issues/892


geigerj:
Actually, let's do this only to GAPIC packages currently consumed by google-cloud-python. Superseded by #150

PR #151, opened 2017-01-06
Add LRO v1

eoogbe:
Uses the new BSD-3-Clause license.


geigerj:
Please either remove the [`gapic-google-longrunning`](https://github.com/googleapis/api-client-staging/tree/master/generated/python/gapic-google-longrunning) directory, or change your directory name so that we don't have duplicate code.

PR #153, opened 2017-01-11
Add grpc streaming tests to pubsub and speech

shinfan:


landrito:
Do these tests run without failure?

shinfan:
@landrito Yes they do

PR #157, opened 2017-02-02
Update GAPIC/proto packages for datastore.

lukesneeringer:
Updated `datastore` autogenerated packages.


lukesneeringer:
@geigerj The datastore and dataflow teams have libraries that they want to ensure are depending on the latest GAPICs, which is the motivation for this. I want you to look over it and make sure everything is okay before I actually publish.

generated/python/gapic-google-cloud-datastore-v1/requirements.txt
```diff
@@ -1,4 +1,4 @@
 googleapis-common-protos>=1.5.0, <2.0dev
-google-gax>=0.15.0, <0.16dev
-grpc-google-cloud-datastore-v1>=0.14.0, <0.15dev
-oauth2client>=2.0.0, <4.0dev
+google-gax>=0.15.4, <0.16dev
+proto-google-cloud-datastore-v1>=0.15.0, <0.16dev
```
bjwatson: Hand-edit to `>=0.90.0, <0.91dev`

generated/python/gapic-google-cloud-datastore-v1/setup.py
```diff
@@ -10,14 +10,14 @@
 
 install_requires = [
     'googleapis-common-protos>=1.5.0, <2.0dev',
-    'google-gax>=0.15.0, <0.16dev',
-    'grpc-google-cloud-datastore-v1>=0.14.0, <0.15dev',
+    'google-gax>=0.15.4, <0.16dev',
+    'proto-google-cloud-datastore-v1[grpc]>=0.15.0, <0.16dev',
```
bjwatson: Hand-edit to `>=0.90.0, <0.91dev`

generated/python/proto-google-cloud-datastore-v1/setup.py
```diff
@@ -11,13 +11,19 @@
...
+    'googleapis-common-protos[grpc]>=1.5.0, <2.0dev',
+    'grpcio>=1.0.2, <2.0dev',
+  ],
+}
+
 setuptools.setup(
-  name='grpc-google-cloud-datastore-v1',
-  version='0.14.0',
+  name='proto-google-cloud-datastore-v1',
+  version='0.15.0',
```
bjwatson: Hand-edit to `0.90.0`

bjwatson: Also, hand-edit `Development Status` below to be `4 - Beta`

generated/python/gapic-google-cloud-datastore-v1/docs/conf.py
```diff
@@ -294,17 +289,15 @@
 # How to display URL addresses: 'footnote', 'no', or 'inline'.
 #texinfo_show_urls = 'footnote'
 
-# If true, do not generate a @detailmenu in the "Top" node's menu.
+# If true, do not generate a @etailmenu in the "Top" node's menu.
```
bjwatson: What happened to the `d` in `@detailmenu`?

generated/python/grpc-google-cloud-datastore-v1/google/cloud/grpc/datastore/v1/datastore_pb2_grpc.py
```diff
@@ -1,157 +0,0 @@
-import grpc
-from grpc.framework.common import cardinality
-from grpc.framework.interfaces.face import utilities as face_utilities
```
bjwatson: Using `git mv` when doing PRs like this with changing package names or code paths can make the diffs easier to read. Just a thought for future such PRs.

generated/python/gapic-google-cloud-datastore-v1/docs/conf.py
geigerj: This is a bug in toolkit; the `@` needs to be escaped because it is a special character in the snippet language. I missed this when copying the template over from packman. Will put up a PR to do this in a minute

geigerj: googleapis/toolkit#981

geigerj:
We don't have handwritten smoke tests (https://github.com/googleapis/api-client-staging/tree/master/test/python) for `datastore` currently. Is it worth adding them?

generated/python/gapic-google-cloud-datastore-v1/docs/conf.py
bjwatson: Thanks @geigerj. I approved your PR.

Seems like we can just hand-edit a fix here, rather than re-generating.

bjwatson:
Good point.

I would say no if `google-cloud-datastore` is integrating with the GAPIC layer soon (so we benefit from their system tests).

If instead we are depending on the Python automated unit tests (which will be done in a few weeks), then we should probably also do a manual smoke test (although we will be automating those a few weeks after finishing unit testing).

generated/python/proto-google-cloud-datastore-v1/setup.py
lukesneeringer: Just to confirm: Are you sure? This hand-edit will continue to be necessary essentially forever (until we automate it in toolkit).

lukesneeringer: (I just did this and pushed, by the way, so it is done unless you decide you want it _un_-done.

bjwatson: As soon as we declare Beta on the Python codegen, we can set all packages to be `0.9x.x` and `4 - Beta`. Main things we need are automated unit and smoke tests, and GAPIC / gRPC docs published for all packages.

We also have talked about allowing version overrides in the GAPIC config, since we may not always want to move the packages together in sync.

lukesneeringer:
@bjwatson Should I push this out to PyPI?

bjwatson:
Yes please

PR #166, opened 2017-02-13
Python Logging GAPIC 0.91.0

lukesneeringer:


lukesneeringer:
Several aspects of this one seem wrong. In particular, the proto- package needs to be 0.90, and also no changes to the GAPIC are detected. Looking into this.

generated/python/gapic-google-cloud-logging-v2/docs/conf.py
```diff
@@ -20,7 +20,7 @@
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 sys.path.insert(0, os.path.abspath('..'))
 
-__version__ = '0.90.0'
+__version__ = '0.15.0'
```
geigerj: This is still wrong. Let's go to `0.91.0`.

lukesneeringer: Missed this one; the rest of it should be `0.91.0`.

generated/python/gapic-google-cloud-logging-v2/google/cloud/gapic/logging/v2/config_service_v2_client.py
```diff
@@ -333,39 +372,56 @@ def update_sink(self,
...
-            logs from one project to a different project. This field is ignored for
-            non-project sinks (e.g. organization sinks) because those sinks are
-            required to have dedicated service accounts.
+          unique_writer_identity (bool): Optional. See
+            `sinks.create <https://cloud.google.com/logging/docs/api/reference/rest/v2/projects.sinks/create>`_
+            for a description of this field.  When updating a sink, the effect of this
+            field on the value of ``writer_identity`` in the updated sink depends on both
+            the old and new values of this field:
+
+            +   If the old and new values of this field are both false or both true,
```
geigerj: I don't think this is being interpreted correctly by the comment fixer (https://github.com/googleapis/toolkit/blob/master/src/main/java/com/google/api/codegen/py/PythonSphinxCommentFixer.java). But let's not block on this until we need to generate docs.

Tracking as https://github.com/googleapis/toolkit/issues/1033.

generated/python/gapic-google-cloud-logging-v2/google/cloud/gapic/logging/v2/logging_service_v2_client.py
```diff
@@ -49,15 +52,17 @@ class LoggingServiceV2Client(object):
...
     _PAGE_DESCRIPTORS = {
         'list_log_entries': _PageDesc('page_token', 'next_page_token',
                                       'entries'),
         'list_monitored_resource_descriptors':
-        _PageDesc('page_token', 'next_page_token', 'resource_descriptors')
+        _PageDesc('page_token', 'next_page_token', 'resource_descriptors'),
+        'list_logs': _PageDesc('page_token', 'next_page_token', 'log_names')
+    }
+
+    _BUNDLE_DESCRIPTORS = {
```
geigerj: @bjwatson @lukesneeringer 
I see that @garrettjonesgoogle added the bundling config back to the GAPIC config for logging a few days ago, resulting in this diff. However, I'm not sure we want to expose bundling on the Python surface given that it's not a feature we maintain or promote. What are your thoughts here?

If we don't want to expose bundling on any GAPIC surface for the time being, we should go ahead and remove this code from the Python GAPIC templates in `toolkit`.

lukesneeringer:
Superceded by #185.

PR #183, opened 2017-02-25
Datastore 0.15.1

lukesneeringer:
A prerequisite step for https://github.com/GoogleCloudPlatform/google-cloud-python/issues/3025

### DIFF
--- a/generated/python/gapic-google-cloud-datastore-v1/requirements.txt
+++ b/generated/python/gapic-google-cloud-datastore-v1/requirements.txt
@@ -1,4 +1,4 @@
 googleapis-common-protos>=1.5.2, <2.0dev
 google-gax>=0.15.6, <0.16dev
-proto-google-cloud-datastore-v1>=0.15.1, <0.16dev
+proto-google-cloud-datastore-v1>=0.90.1, <0.91dev
 oauth2client>=2.0.0, <4.0dev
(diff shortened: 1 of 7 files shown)


lukesneeringer:
I figured out while looking at this PR why this one was special/different; I had forgotten that we released an 0.15.0 for this several weeks earlier. I must not have noticed when firing through commands to release all the new ones later. (Notice how this one does not have the GAPIC `lib_name` changes.)

generated/python/gapic-google-cloud-datastore-v1/requirements.txt
```diff
@@ -1,4 +1,4 @@
-googleapis-common-protos>=1.5.0, <2.0dev
-google-gax>=0.15.4, <0.16dev
-proto-google-cloud-datastore-v1>=0.90.0, <0.91dev
```
bjwatson: Manually fix the `proto-` version

generated/python/gapic-google-cloud-datastore-v1/setup.py
```diff
@@ -9,15 +9,15 @@
 import sys
 
 install_requires = [
-    'googleapis-common-protos>=1.5.0, <2.0dev',
-    'google-gax>=0.15.4, <0.16dev',
-    'proto-google-cloud-datastore-v1[grpc]>=0.90.0, <0.91dev',
```
bjwatson: Ditto.

generated/python/gapic-google-cloud-datastore-v1/requirements.txt
bjwatson: Also, shouldn't there be a `[grpc]` option here?

PR #187, opened 2017-03-02
Refresh java clients with latest toolkit

shinfan:
- Update pubsub grpc to 0.1.6 due to proto changes


shinfan:
I will sync the api-client-staging with google-cloud-java after this PR is merged.

generated/java/build.gradle
```diff
@@ -1,6 +1,7 @@
 import groovy.io.FileType
 
 def generated_grpc_version = "0.1.5"
+def pubsub_grpc_version = "0.1.6"
```
garrettjonesgoogle: call this `generated_pubsub_grpc_version`.

More generally: I would like to get into the habit of generating the grpc packages all at the same version. It is much easier to manage refreshes if we do that. This change is a case in point: you have to do something special for pubsub because its version is different than the rest.

generated/java/google-cloud-pubsub-v1/src/main/java/com/google/cloud/pubsub/spi/v1/SubscriberClient.java
```diff
@@ -1030,6 +1109,339 @@ public final void modifyPushConfig(ModifyPushConfigRequest request) {
...
+   *     `projects/{project}/snapshots/{snap}`.
+   * @param subscription The subscription whose backlog the snapshot retains. Specifically, the
+   *     created snapshot is guaranteed to retain: (a) The existing backlog on the subscription.
+   *     More precisely, this is defined as the messages in the subscription's backlog that are
+   *     unacknowledged upon the successful completion of the `CreateSnapshot` request; as well as:
+   *     (b) Any messages published to the subscription's topic following the successful completion
+   *     of the CreateSnapshot request. Format is `projects/{project}/subscriptions/{sub}`.
+   * @throws com.google.api.gax.grpc.ApiException if the remote call fails
+   */
+  public final Snapshot createSnapshot(String name, String subscription) {
```
garrettjonesgoogle: We need to fix the configuration for pubsub so that we take `SubscriptionName` here.

For expediency, maybe we should skip the pubsub update in this round of updates, because the header updates are time sensitive.

generated/java/google-cloud-pubsub-v1/src/main/java/com/google/cloud/pubsub/spi/v1/SubscriberClient.java
```diff
@@ -1030,6 +1109,339 @@ public final void modifyPushConfig(ModifyPushConfigRequest request) {
...
+   *   String formattedSnapshot = SnapshotName.create("[PROJECT]", "[SNAPSHOT]").toString();
+   *   subscriberClient.deleteSnapshot(formattedSnapshot);
+   * }
+   * </code></pre>
+   *
+   * @param snapshot The name of the snapshot to delete. Format is
+   *     `projects/{project}/snapshots/{snap}`.
+   * @throws com.google.api.gax.grpc.ApiException if the remote call fails
+   */
+  public final void deleteSnapshot(String snapshot) {
```
garrettjonesgoogle: We need a resource name class for snapshot names applied here.

garrettjonesgoogle:
We still want the header updates to be applied to pubsub... Can you work off of an older pubsub.proto + pubsub_gapic.yaml?

shinfan:
@garrettjonesgoogle Skipped pubsub in the PR. PTAL

shinfan:
@garrettjonesgoogle  Added pubsub with previous proto + gapic_config. PTAL

garrettjonesgoogle:
LGTM

PR #191, opened 2017-03-03
Python Datastore GAPIC.

lukesneeringer:
This will likely be the highest risk of the new autogen GAPICs because it has hand-edits.


generated/python/gapic-google-cloud-datastore-v1/google/cloud/gapic/datastore/v1/datastore_client.py
```diff
@@ -321,6 +332,11 @@ def commit(self,
           :exc:`google.gax.errors.GaxError` if the RPC is aborted.
           :exc:`ValueError` if the parameters are invalid.
         """
+        # Sanity check: We have some fields which are mutually exclusive;
+        # raise ValueError if more than one is sent.
+        oneof.check_oneof(transaction=transaction, )
```
lukesneeringer: I am curious how this happened. (It is harmless, but how is there a one-of with only one possibility?)

PR #196, opened 2017-03-03
Python PubSub GAPIC 0.15.3

lukesneeringer:


generated/python/gapic-google-cloud-pubsub-v1/google/cloud/gapic/pubsub/v1/publisher_client.py
```diff
@@ -213,8 +204,7 @@ def __init__(self,
             client_config,
             config.STATUS_CODE_NAMES,
             metrics_headers=metrics_headers,
-            bundle_descriptors=self._BUNDLE_DESCRIPTORS,
```
geigerj: _Technically_ this is a breaking change, so by convention we'd bump to `0.16.0`. In practice, I'm fairly sure no one is using GAPIC bundling, which is a complicated and unpublicized feature. What are your thoughts here?

lukesneeringer: I could go either way, but it is kind of important to me to keep everything together. So, if this goes to 0.16, they all do.

That would mean that we do not get the benefit of semi-automatic opt-in on metrics without bumping all the Veneers.

I think because the use case is uncommon, un-publicized, and does not break anything in the Veneer, that we should let it ride. But very open to a conflicting opinion.

geigerj: That's convincing enough to me, especially since SemVer technically doesn't make any such guarantees about breaking patch changes in version 0.

PR #204, opened 2017-03-07
Regenerate spanner clients.

jmuk:
This has several manual updates actually. We should fix those
manual efforts eventually;
- streaming methods with errors test cases fail. This is because
  of nodejs behavior where delivery of 'data' events are queued
  but unhandled 'error' events actually throw the error immediately.
  See: https://github.com/googleapis/toolkit/issues/1108
- the path name and module name confusion for spanner admin
  APIs.


jmuk:
Split the things. Also I've found a different approach of avoiding the error; the manual edits are quite small now.

PR #207, opened 2017-03-15
Generate cloud debugger gapic

swcloud:


hxiong388:
I forgot to mention the [debugger gRPC](https://github.com/googleapis/api-client-staging/blob/b4c55b498a7295bbb67ea4431744845510723ba6/generated/ruby/google-cloud-ruby/google-cloud-debugger/lib/google/devtools/clouddebugger/v2/data_pb.rb#L83-L84) also depends on google.devtools.source.v1.source_context. This GAPIC client is not complete without it.

PR #208, opened 2017-03-15
Regenerate Java with bundling closeable removed

michaelbausor:


garrettjonesgoogle:
You need to update the GAX dependency, right?

garrettjonesgoogle:
LGTM

PR #211, opened 2017-03-28
Refresh Java staging clients with latest toolkit

shinfan:
Update to GAX 0.7.0 with the `PropertiesProvider` Change.

### DIFF
--- a/generated/java/build.gradle
+++ b/generated/java/build.gradle
@@ -2,6 +2,9 @@ import groovy.io.FileType
 
 def generated_grpc_version = "0.1.5"
 
+// TODO: Update all grpc versions to 0.1.6
+def pubsub_grpc_version = "0.1.6"
+
 buildscript {
   repositories {
     mavenLocal()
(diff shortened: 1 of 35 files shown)


garrettjonesgoogle:
Could we update PubSub and the header version separately from each other? @michaelbausor and @neozwu were working on updating PubSub.

michaelbausor:
@garrettjonesgoogle We got to the point of regenerating core protos, but found that there was an error with the packman generation (missing some protos). Rather than invest effort to debug and fix, we discussed with @shinfan and decided to wait until https://github.com/googleapis/artman/pull/143 is merged, and then can regenerate without packman.

michaelbausor:
We could submit the pubsub regeneration with its own version bump but without republishing all other proto packages as @shinfan has done here, except it is own PR.

garrettjonesgoogle:
I think the 0.1.6 version of grpc-google-cloud-pubsub-v1 might have been generated from the latest protos already.

shinfan:
Conflicts resolved. PTAL

garrettjonesgoogle:
LGTM

michaelbausor:
LGTM

PR #215, opened 2017-03-31
Speech v1 GAPIC.

lukesneeringer:


shinfan:
LGTM, let @geigerj approve as well.

lukesneeringer:
Just merged googleapis/toolkit#1152 so we could get tests also. :-)

generated/python/gapic-google-cloud-speech-v1/requirements.txt
```diff
@@ -0,0 +1,4 @@
+googleapis-common-protos>=1.5.2, <2.0dev
+google-gax>=0.15.7, <0.16dev
```
geigerj: 0.15.8?

lukesneeringer: The changes in 0.15.8 did not require upgrades from downstream clients.

PR #214, opened 2017-03-31
java: update generated code

pongad:
reflects googleapis/toolkit#1153 .

PR #216, opened 2017-03-31
Java regeneration: PartitionKey, @ExperimentalApi

garrettjonesgoogle:
Partially supersedes https://github.com/googleapis/api-client-staging/pull/214 (I'm not regenerating Speech)

### DIFF
--- a/generated/java/build.gradle
+++ b/generated/java/build.gradle
@@ -32,7 +32,7 @@ subprojects {
   ext {
     // Shortcuts for libraries we are using
     libraries = [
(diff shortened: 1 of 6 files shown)


garrettjonesgoogle:
I realized a flaw: Only the client classes are getting marked with `@ExperimentalApi`; there are more classes that need to be marked. I will do another update in toolkit.

garrettjonesgoogle:
Hmm well it turns out that the other classes are hardcoded to `@ExperimentalApi` in toolkit, so it had never been removed. Looks like the net result is that the code is in the state I want it.

PR #218, opened 2017-04-05
NodeJS Natural Language v1beta2 GAPIC

lukesneeringer:


geigerj:
Otherwise, looks good.

lukesneeringer:
> Seems a little counterintuitive having v1beta1 under the language-v1 directory? Is the v1beta1 code still used in google-cloud-node at all?

I was wondering about that too, as something I wanted to ask about later.
Node only uses a subset of the actual generated code.

lukesneeringer:
My guess is that is correct also.

PR #219, opened 2017-04-06
(DO NOT MERGE) Java: generating nonfunctional compute client

garrettjonesgoogle:
This isn't for review exactly, it's more to look at for amusement.


bjwatson:
@garrettjonesgoogle How did you create this? Is everything in `google-cloud-compute-v1/src/main/java/com/google/cloud/compute/spi/v1` hand-written using an existing GAPIC template, and then everything else generated with `protoc` from the proto2 files?

garrettjonesgoogle:
@bjwatson I hand-converted the proto 2 file for InstancesService to proto 3 (and put that in compute_instances.proto), hand-converted the service yaml, used the gapic config generator to create the gapic yaml, then generated the client.

bjwatson:
Ok, that makes sense.

PR #222, opened 2017-04-11
Regenerate with changes to page

michaelbausor:
Also:
- Adds speech v1
- Moves language v1beta2 to google-cloud-language-v1beta2 folder
- Updates longrunning


generated/java/google-cloud-logging-v2/src/main/java/com/google/cloud/logging/spi/v2/ConfigClient.java
```diff
@@ -95,7 +94,6 @@
  * </pre>
  */
 @Generated("by GAPIC")
-@ExperimentalApi
```
garrettjonesgoogle: You're missing something here (toolkit change or googleapis change) - this was just recently added in.

garrettjonesgoogle:
LGTM

PR #223, opened 2017-04-13
Refresh Java gapic clients with updated property path

shinfan:
- Also updated the polling interval for Speech v1

### DIFF
--- a/generated/java/build.gradle
+++ b/generated/java/build.gradle
@@ -1,8 +1,6 @@
 import groovy.io.FileType
 
-def generated_grpc_version = "0.1.5"
-def generated_language_version = "0.1.6"
-def generated_pubsub_version = "0.1.6"
+def generated_grpc_version = "0.1.7"
 
 buildscript {
   repositories {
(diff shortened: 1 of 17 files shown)


generated/java/build.gradle
```diff
@@ -124,14 +122,14 @@ project(':generated:java:google-cloud-logging-v2') {
 
 project(':generated:java:google-cloud-pubsub-v1') {
   dependencies {
-    compile 'com.google.api.grpc:grpc-google-cloud-pubsub-v1:' + generated_pubsub_version
+    compile 'com.google.api.grpc:grpc-google-cloud-pubsub-v1:' + generated_grpc_version
   }
 }
 
 project(':generated:java:google-cloud-language-v1') {
```
michaelbausor: Speech and language have their versions in separate packages; we can fix in this PR or submit separately.

garrettjonesgoogle:
LGTM

generated/java/build.gradle
shinfan: Sorry can you be more specific about the issue? I am not sure what you are referring to.

michaelbausor: We need separate `project` declarations for the different versions of an API, because they are in different folders. I think this is better in a separate PR.

shinfan: Okay I see.

PR #232, opened 2017-04-27
Python CI.

lukesneeringer:
This commit adds a CI setup for Python. WIP.

_In its current state, the actual tests should fail_, because the existing code on master right now has failing tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,8 +37,9 @@ jobs:
           command: pip install nox-automation
       - run:
           name: Run GAPIC unit tests.
-          command: nox --noxfile generated/python/nox.py \
-                       --sessions "unit_tests(python_version='2.7')"
+          command: |
+            nox --noxfile generated/python/nox.py \
+                --sessions "unit_tests(python_version='2.7')"
     working_directory: /var/code/googleapis/
 
   python34:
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -3,9 +3,86 @@ version: 2
 jobs:
   build:
     docker:
-      - image: ubuntu:16.04
+      - image: googleapis/curl:latest
```
geigerj: What/where is this image?

.circleci/config.yml
```diff
@@ -3,9 +3,86 @@ version: 2
...
     steps:
       - run:
-          name: No-op.
-          command: echo "This command did nothing. It exists to configure CircleCI so that PRs that supply actual configuration will work."
-    working_directory: /var/code/googleapis
+          name: Add command to be able to trigger jobs.
+          command: |
+            touch trigger_job.sh && chmod a+x trigger_job.sh
```
geigerj: Why are you creating/storing this as a file instead of just defining a function?

lukesneeringer:
Quick brain dump of our plan here:

#### Status Quo

Right now, all but three APIs do not have tests, but three (pubsub, speech, language) were committed while work on the tests was in progress, and they fail.
  
#### The Plan

After discussing with @geigerj, what we intend to do is:

  * Merge this. CI will intentionally go red.
  * Work on #233, #234, and #235, in order. For each we will show and document that the appropriate tests went from red to green. Whichever one we do last should be a green build.
  * Bump all the remaining GAPICs, with proper CI verification for each.

.circleci/config.yml
lukesneeringer: Every command is independent; the function definition would not survive to the next step.

.circleci/config.yml
lukesneeringer: On Docker Hub. It is two lines.

lukesneeringer: If you are curious, the entire Dockerfile is:

```Dockerfile
FROM alpine:latest
RUN apk add --update curl \
  && rm -rf /var/cache/apk/*
```

PR #238, opened 2017-04-27
Python GAPIC: datastore v1

lukesneeringer:
This pull request was generated by artman. Please review it thoroughly before merging.


lukesneeringer:
This is broken because the version number is wrong. Will fix after lunch.

PR #256, opened 2017-04-28
One-way auto-sync to private.

lukesneeringer:
This commit makes it such that successful merges to master will also update the private mirror.

/cc @jskeet


.circleci/config.yml
```diff
@@ -86,3 +92,14 @@ jobs:
...
+
+  sync_to_private:
+    docker:
+      - image: googleapis/git
+    steps:
+      - checkout
+      - run:
+          command: |
+            git remote add private https://googleapis-publisher:${GITHUB_TOKEN}@github.com/googleapis/api-client-staging-private.git
+            git push private HEAD:master 5a648c2b119da694dd7da26adf7443e0660d6caa
```
geigerj: I am not familiar with this syntax. What does the hash `5a64...` do?

lukesneeringer: That is a copy-paste failure. Will fix (and regen that token :-)).

PR #305, opened 2017-05-17
[WIP] Testing CircleCI workflows alpha.

lukesneeringer:
CircleCI was kind enough to ask us to alpha test their new workflows feature, which allows us to do the configuration we want in `api-client-staging` without my janky shell script (and the risk of exposing a CircleCI API token).

This is an attempt to see how that works and what it does. It is unlikely to work on the first attempt so consider this to be WIP.


lukesneeringer:
This is now working. Thanks to the CircleCI engineers who are improving their multi-image workflow. :-)

@shinfan Please review. (Note that the OpenJDK 7 failure is a separate issue.)

PR #307, opened 2017-05-18
WIP: vendor generated Go clients

pongad:
We want to be able to test generated clients from this repo.
However, we added import comment saying these clients belong in
"cloud.google.com/go".
Work around this by putting all clients in vendor directory,
which disables the canonical import check.

Packages internal/version and internal/testutil are redeclared
in this repo.
Package testutil should later be changed to match how we plumb
project ID and credentials through CircleCI.

Some tests do not build yet, since they depend on testutil.UIDSpace.
This should be fixed by googleapis/toolkit#1279.

If this new layout looks good to reviewers,
a corresponding change to be made to googleapis to preserve the layout.
Hence, the WIP in the commit title.


generated/go/google-cloud-speech-v1beta1/cloud.google.com/go/speech/apiv1beta1/doc.go
```diff
@@ -1,33 +0,0 @@
-// Copyright 2017, Google Inc. All rights reserved.
```
pongad: Note to reviewers: these deleted files are duplicates and can be safely deleted.

pongad:
I'm sorry @jba, please take another look.

Copied from commit message:
> The previous attempt satisfies Go but not artman.
> Artman deletes the contents of a directory before placing
> generated files in it.
>
> For this reason, we cannot put all APIs in the same directory,
> lest they clobber each other.

The structure looks awful but everything seems to work.

PR #310, opened 2017-05-18
Node.js testing

lukesneeringer:
This PR adds testing in `api-client-staging` for Node.js GAPICs.

Caveats:

  * It removes `language-v1beta2` because of a collision.
  * It removes `dlp` because the protos are not in the common protos package.
  * It only tests on Node 6 and Node 7. The tests fail on Node 4.

The latter two at minimum needs to be re-added in a new PR, and tests made to pass. However, this is _still_ better than the _status quo_ (which is not running the tests at all).


.circleci/config.yml
```diff
@@ -19,17 +19,23 @@ jobs:
...
       - run:
-          name: 'Trigger jobs: Python Unit Tests'
+          name: 'Trigger jobs: Node.js Unit Tests'
           command: |
-            ./trigger_job.sh python27
-            ./trigger_job.sh python34
-            ./trigger_job.sh python35
-            ./trigger_job.sh python36
+            ./trigger_job.sh node4
+            ./trigger_job.sh node6
```
shinfan: Is there any reason that we does not test node5?

generated/nodejs/iam/src/index.js
```diff
@@ -0,0 +1,51 @@
...
+  opts.libName = 'gccl';
+  opts.libVersion = VERSION;
+
+  // Create the image annotator client with the provided options.
+  var client = gapic.v1(opts).iamClient(opts);
+  return client;
+}
+
+module.exports = iam_v1;
+module.exports.v1 = iam_v1;
```
shinfan: Add newline

lukesneeringer: This is auto-generated code.

lukesneeringer:
> Does it mean that we have compatibility issues with Node 4?

I am not sure. I am investigating that now. My guess is no, since I think we would know about it, but I am not certain.

shinfan:
LGTM, I am okay with merging this for now and investigate the Node4 issue after. (I have created an issue to track https://github.com/googleapis/api-client-staging/issues/311)

landrito:
@lukesneeringer I'm curious about why was the path changed for all the nodejs files? This caused an inconsistency between googleapis artman yamls and this directory making artman generation not work for nodejs.

Based on your first commit I assume it's because having google-cloud-node is arbitrary. I think the original reason for having it generate to google-cloud-node/packages is to have similar file structure to google-cloud-node.

Do you suggest that we move these back to g-c-n/packages or to change the artman yamls in googleapis.

PR #314, opened 2017-05-22
add Go CI

pongad:


.circleci/config.yml
```diff
@@ -184,6 +188,16 @@ jobs:
                 --sessions "unit_tests(python_version='3.6')"
     working_directory: /var/code/googleapis/
 
+  go18:
+    docker:
+      - image: circleci/golang:1.8.1
```
lukesneeringer: Can we use `golang:1.8` instead?

lukesneeringer:
With #305 merged, this is now out of date; I am going to update it.

pongad:
@lukesneeringer The build seems to be failing with "job 'build' is not found". Could this be caused by circleci running on pongad/api-client-staging instead of googleapis/api-client-staging? I admittedly don't know enough about circleci to know what's wrong.

lukesneeringer:
Yep, that is a bug with CircleCI's (brand new) workflow process.
Feel free to make a PR from the primary origin to get this in. I am coordinating with CircleCI to get the root problem fixed.

PR #315, opened 2017-05-23
NodeJS: Move the packages to a folder that matches the folder in GCN

landrito:
All of the artman configs specify for the generated files to be moved to the directory `generated/nodejs/google-cloud-node/packages/{shortname}`. This changes api-client-staging to match that.


lukesneeringer:
Can we change the config, rather than this? It is annoying to have this go so deeply nested for no obvious reason.

landrito:
SG

PR #316, opened 2017-05-23
Set openjdk7 to use a specific version

shinfan:
The latest openJDK7 (131) image introduces a SSL security issue. (Already reported [here](https://github.com/docker-library/openjdk/issues/117))

An older version solves the test failures. We can revert to use the latest image once the issue is resolved at their end.


garrettjonesgoogle:
LGTM

PR #318, opened 2017-05-23
Sync common protos packages

shinfan:


shinfan:
@garrettjonesgoogle  I have removed servicemanager/control PTAL

generated/java/proto-google-common-protos/src/main/proto/google/api/servicemanagement/v1/servicemanager.proto
```diff
@@ -1,4 +1,4 @@
-// Copyright 2016 Google Inc.
+// Copyright 2017 Google Inc.
```
garrettjonesgoogle: Why is the proto still included?

shinfan: Done

shinfan:
@garrettjonesgoogle PTAL

garrettjonesgoogle:
LGTM

PR #332, opened 2017-06-03
Update build files to support batch javadoc update

shinfan:
- Switch to use absolute paths in the build.gradle
- Set `projectDir` for each submodule in the root settings

After this PR you can run to batch update the javadoc for all gRPC/Proto packages
```
./gradlew updateAndPushDocs
```

cc/ @lukesneeringer


garrettjonesgoogle:
LGTM

PR #349, opened 2017-06-15
Regenerate PHP with documentation updates

michaelbausor:
Toolkit PR: https://github.com/googleapis/toolkit/pull/1361

- Adds `@experimental` annotation to client class and methods
- Escape only `*/` exactly instead of `*` (#1278)
- Update generated param doc for enums (#1277)
- Remove class comment when no resource format method present (#1276)

cc @dwsupplee @jdpedrie

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,7 +1,7 @@
 {
     "name": "google/api-client-staging",
     "require": {
-        "google/gax": "^0.9",
+        "google/gax": "^0.10",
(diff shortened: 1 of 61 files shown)


dwsupplee:
Awesome!

One small item: For the doc site, in order to parse `@see` tags properly, we need the FQNS, not just the relative.

michaelbausor:
Thanks, updated with FQNS

PR #361, opened 2017-06-23
java: update build.gradle and add video-intelligence

pongad:
The build.gradle change reflects googleapis/toolkit#1382.

Generated Code:
Dependency changes are described in GoogleCloudPlatform/google-cloud-java#2171

Changes to Java code
- removing retry to UNAVAILABLE described in GoogleCloudPlatform/google-cloud-java#2171
- LRO change described in GoogleCloudPlatform/google-cloud-java#2164.

The rest are changes to generated protobuf classes introduced by updating `protoc` to v3.3.0

PR #362, opened 2017-06-26
java: update gax dependency, making tests pass

pongad:
gax dependency is updated to 1.4.1, following googleapis/gax-java#332.

google-longrunning-v1 now has its own build.gradle,
since its dependencies were removed
and inadvertently not replaced
in #361.


garrettjonesgoogle:
LGTM

PR #364, opened 2017-06-28
Refresh php, add protobuf types

michaelbausor:


shinfan:
Could you provide more information about how this PR affect the release process of PHP such as:
- Is now the PHP clients self-contained after this PR?
- Will the protobuf types be synced into google-cloud-php? If we publish these protobuf types as separate package, can the process be automated based on the current structure?
- Will the CI tests depend on the local proto types?

michaelbausor:
> - Is now the PHP clients self-contained after this PR?

Yes, the PHP clients are now self contained
> - Will the protobuf types be synced into google-cloud-php? If we publish these protobuf types as separate package, can the process be automated based on the current structure?

Eventually yes, but currently no. Right now we are still using [proto-client-php](https://github.com/googleapis/proto-client-php) to hold the protobuf types. Once the php_namespace option is included in a protobuf release, we will be able to distribute the protobuf types in google-cloud-php. I think that automating the process based on this structure would be possible, but probably not desirable given that it will not be used after we can make the change to distributing in google-cloud-php
> - Will the CI tests depend on the local proto types?

I had not fully considered this, but yes that can be made to work. I have updated to PR to remove the dependency on `proto-client-php`, to ensure that the protobuf types used will be the ones distributed locally.

shinfan:
I plan to add CI for PHP staging soon (probably next week)

This can be done in a separate PR but I would recommend PHP to stage common-protos (like Java) so all the proto dependencies are locally staged for testing.

michaelbausor:
That makes sense - I will add that in a followup PR. It should only be changes in the googleapis settings, and an addition to the composer.json file

PR #369, opened 2017-07-06
Java: regenerate DLP proto and grpc

neozwu:
This PR regenerates DLP proto and grpc version 0.1.13, which was missed in the previous batch generation. It also applies a few changes introduced by protoc 3.3.0


garrettjonesgoogle:
LGTM

PR #371, opened 2017-07-11
Add CI for PHP 5

shinfan:
Add PHP5 first. The unit tests are now having compatibility issues with PHP 7.*.
Note: A testing Docker image is built and pushed to Dockerhub:  @https://hub.docker.com/r/googleapis/php

TODO: Fix issues on PHP 7.

cc @landrito

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -163,6 +163,19 @@ jobs:
                 --sessions "unit_tests(python_version='3.6')"
     working_directory: /var/code/googleapis/
 
+  php5:
+    docker:
+      - image: googleapis/php:5
+    steps:
+      - checkout
+      - run:
+          name: Install PHP library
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -163,6 +163,19 @@ jobs:
                 --sessions "unit_tests(python_version='3.6')"
     working_directory: /var/code/googleapis/
 
+  php5:
```
michaelbausor: This may come later when we add more platforms, but we may want to call this `php55` or `php56` depending on which version of PHP is being used

shinfan: Done. Changed to 5.6

shinfan:
@lukesneeringer ci/circleci:sync failed. Is this expected?

Issue #373, opened 2017-07-12
PHP unit tests crashed on PHP 7.*


shinfan:
Error message:
root@c206340828e2:/api-client-staging# ./vendor/bin/phpunit 

Fatal error: Class 'PHPUnit_Framework_TestCase' not found in /api-client-staging/vendor/google/gax/src/Testing/GeneratedTest.php on line 36

michaelbausor:
Fixed in https://github.com/googleapis/api-client-staging/pull/375

PR #375, opened 2017-07-13
Update composer to support PHP 7

michaelbausor:
- Fix version of PHPUnit
- Bump GAX to version with fix: https://github.com/googleapis/gax-php/pull/85


michaelbausor:
Fixes https://github.com/googleapis/api-client-staging/issues/373

PR #380, opened 2017-07-24
Refresh PHP proto types

michaelbausor:
Updates to documentation only

### DIFF
--- a/generated/php/google-cloud-error-reporting-v1beta1/proto/src/Google/Devtools/Clouderrorreporting/V1beta1/DeleteEventsRequest.php
+++ b/generated/php/google-cloud-error-reporting-v1beta1/proto/src/Google/Devtools/Clouderrorreporting/V1beta1/DeleteEventsRequest.php
@@ -11,7 +11,7 @@ use Google\Protobuf\Internal\GPBUtil;
 /**
  * Deletes all events in the project.
  *
- * Protobuf type <code>Google\Devtools\Clouderrorreporting\V1beta1\DeleteEventsRequest</code>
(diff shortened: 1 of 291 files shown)


generated/php/google-cloud-error-reporting-v1beta1/proto/src/Google/Devtools/Clouderrorreporting/V1beta1/DeleteEventsRequest.php
```diff
@@ -11,7 +11,7 @@
 /**
  * Deletes all events in the project.
  *
- * Protobuf type <code>Google\Devtools\Clouderrorreporting\V1beta1\DeleteEventsRequest</code>
+ * Generated from protobuf message <code>google.devtools.clouderrorreporting.v1beta1.DeleteEventsRequest</code>
```
shinfan: Why do you change the package path here? Isn't the path separated by "\" in PHP?

michaelbausor: Yes, but this is referring to the protobuf message type, which is lowercase and `.` separated. This is a change in PHP protoc which we are now picking up in our generated proto classes.

shinfan: Okay that make sense

shinfan:
LGTM

PR #382, opened 2017-07-27
Refresh Java

shinfan:


michaelbausor:
It looks like the change to add the `optional` string is coming because a different version of protoc was used than the last release. Is this what we want?

garrettjonesgoogle:
I think that's a regression.

shinfan:
Fixed. PTAL.

michaelbausor:
LGTM

PR #384, opened 2017-07-28
Refresh PHP from googleapis, except with pubsub gapic yaml reverted

michaelbausor:


garrettjonesgoogle:
The commit description would be more accurate as "Refresh PHP from googleapis, except with pubsub reverted". Otherwise, LGTM

PR #400, opened 2017-08-18
Refresh Java clients

shinfan:
Upgrade GAX dependency to 1.6.0 and 0.23.0


generated/java/gapic-google-cloud-dlp-v2beta1/src/main/java/com/google/cloud/dlp/v2beta1/DlpServiceClient.java
```diff
@@ -483,7 +503,7 @@ public final RedactContentResponse redactContent(RedactContentRequest request) {
    *
    * @param name Identifier of the results set returned as metadata of the longrunning operation
    *     created by a call to CreateInspectOperation. Should be in the format of
-   *     `inspect/results/{id}.
+   *     `inspect/results/{id}`.
```
garrettjonesgoogle: what caused this the second tick to be lost here?

michaelbausor: This was missing from the proto and fixed upstream

garrettjonesgoogle:
LGTM

PR #403, opened 2017-09-06
Java: Add logging-external-types

eoogbe:
Updates googleapis/googleapis#187


geigerj:
This LGTM but @garrettjonesgoogle should give the "real" review; I'm not very familiar with the expected structure of the Java package.

Also, could you add "Updates googleapis/googleapis#187" to the descriptions of your PRs? It adds context and will make it easier to cross reference them in the future.

garrettjonesgoogle:
I'm going to have @michaelbausor review in my stead

michaelbausor:
It looks like we need to include appengine.proto in the gapic yaml (https://github.com/googleapis/googleapis/blob/master/gapic/core/artman_logging_external_types.yaml#L12), since audit_data.proto depends on it for the `UpdateServiceRequest` (and possibly more, I only investigated that far)

eoogbe:
If we follow the dependencies, we need to include the whole `google/appengine/v1` path.

PR #404, opened 2017-09-06
Java: Add appengine-v1 proto package

eoogbe:
Updates googleapis/googleapis#187


geigerj:
LGTM, but @michaelbausor should approve

michaelbausor:
When I run batch generation, I also get a grpc-google-appengine package, which fails to compile. Is this something unique to me, or are you also seeing this? And given that we are generating it, should we also be publishing it to api-client-staging?

eoogbe:
@michaelbausor What is batch generation?

michaelbausor:
Batch generation is using the `--batch` flag with artman instead of specifying a particular API. This generates all APIs. If we only want to generate a proto package, and not a package that contains the gRPC stubs, we could update the artman yaml to remove the `grpc` target: https://github.com/googleapis/googleapis/blob/master/gapic/core/artman_logging_external_types.yaml#L31

eoogbe:
`artman --batch --lang java --pipeline ProtoClientPipeline` fails with `ValueError: src_proto_path must be provided` when `artman_logging_external_types.yaml` unchanged.

eoogbe:
Ok, you're not supposed to specify a pipeline. We should make that invalid with the batch argument.

If I can't specify the  ProtoClientPipeline, it generates an empty grpc package. But I can't remove the grpc path because ProtoClientPipeline still depends on it for some reason.

Going to exclude appengine from the batch apis.

eoogbe:
@michaelbausor PTAL

generated/java/proto-google-appengine/build.gradle
```diff
@@ -0,0 +1,168 @@
...
+  }
+}
+
+apply plugin: 'java'
+apply plugin: 'maven'
+apply plugin: 'signing'
+
+description = 'PROTO library for proto-google-appengine-v1'
+group = "com.google.api.grpc"
+version = "0.1.15"
```
michaelbausor: This should be 0.1.18. The required updates in googleapis to ensure 0.1.18 is generated are in this PR: https://github.com/googleapis/googleapis/pull/455 but the googleapis changes need to be made upstream - there is an internal change pending.

eoogbe:
@garrettjonesgoogle Requesting review from you since Michael is OOO

PR #406, opened 2017-09-11
Add googleapis-common-protos

geigerj:
This differs from the published googleapis-common-protos 1.3.5 only
in that it was generated using a newer version of
`error_details.proto`


geigerj:
@landrito This only adds (aside from the one diff noted in the commit message) the current state of affairs, so that there will be a meaningful baseline from which to compare the diffs I'm about to make.

geigerj:
Superseded by #408

PR #408, opened 2017-09-11
Split non-gRPC types from googleapis-common-protos

geigerj:


geigerj:
Note: I will not squash these two commits to preserve the diff.

geigerj:
Note: CI failures are expected because the googleapis-common-protos-types gem is not yet on rubygems.

geigerj:
@landrito Sure, I can do that (so hopefully CI will pass on the merged commits).

PR #416, opened 2017-09-27
Java Billing proto update

andreamlin:
billing.proto was updated in googleapis in [this commit](https://github.com/googleapis/googleapis/commit/9f5a5eca1e482b8cc4f534e7c2e4ff27634cc6ee) and the proto and grpc libraries need to be updated. 

The version numbers are bumped to 0.1.20 accordingly.


generated/java/proto-google-common-protos/src/main/java/com/google/api/Advice.java
```diff
@@ -182,17 +182,6 @@ public int hashCode() {
...
-      throws com.google.protobuf.InvalidProtocolBufferException {
-    return PARSER.parseFrom(data);
-  }
-  public static com.google.api.Advice parseFrom(
-      java.nio.ByteBuffer data,
-      com.google.protobuf.ExtensionRegistryLite extensionRegistry)
-      throws com.google.protobuf.InvalidProtocolBufferException {
-    return PARSER.parseFrom(data, extensionRegistry);
-  }
-  public static com.google.api.Advice parseFrom(
```
garrettjonesgoogle: Something seems wrong if the generated code is having public methods deleted

andreamlin: If I run "protoc google/api/config_change.proto", I get the same result as above. It might be my outdated protoc (v3.2) instead of the newest version 3.4.0. I'll update and see if that fixes it.

andreamlin:
I signed it!

generated/java/proto-google-common-protos/src/main/java/com/google/api/Advice.java
andreamlin: Ok, everything's been regenerated using 3.4.0 (as specified in https://github.com/googleapis/toolkit/blob/master/build.gradle:52).

The previous version of this java file was compiled using protoc v3.2.0 (toolkit's protoc version bumped from [3.2 to 3.4 on 9/5/17](https://github.com/googleapis/toolkit/commit/0d0761fa7bb22f0deacf3f0732f1c27bd9a6e9b2#diff-c197962302397baf3a4cc36463dce5ea), but this file last updated [June 2017](https://github.com/googleapis/api-client-staging/commits/master/generated/java/proto-google-common-protos/src/main/java/com/google/api/Advice.java)), so that could explain why there are so many changes in the java compiled proto classes.

andreamlin:
PTAL.

generated/java/proto-google-common-protos/src/main/java/com/google/api/Advice.java
andreamlin: Actually going to recompile protos with version 3.3, as specified in artman's Dockerfile.

garrettjonesgoogle:
My concern is addressed - @vam-google or @neozwu can do final approval.

andreamlin:
PTAL.

PR #420, opened 2017-10-22
Add bigtable GAPIC

andreamlin:
Adding generated directories:
- generated/java/gapic-google-cloud-bigtable-v2
- generated/java/gapic-google-cloud-bigtable-admin-v2
- generated/java/proto-google-cloud-bigtable-v2
- generated/java/proto-google-cloud-bigtable-admin-v2
- generated/java/grpc-google-cloud-bigtable-v2
- generated/java/grpc-google-cloud-bigtable-admin-v2


andreamlin:
PTAL
I'm not sure why there's a golang test failure in DLP, given that no DLP files were changed/added/deleted.
The circle CI java test failure is due to a "WARNING" being thrown, but it isn't a test "ERROR," so that should be ok.

andreamlin:
Reran CI and it's all green now!

garrettjonesgoogle:
cc @igorbernstein2

generated/java/gapic-google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/v2/BigtableClient.java
```diff
@@ -0,0 +1,636 @@
...
+ * </code>
+ * </pre>
+ */
+@Generated("by GAPIC v0.0.5")
+@BetaApi
+public class BigtableClient implements BackgroundResource {
+  private final BigtableSettings settings;
+  private final BigtableStub stub;
+
+  private static final PathTemplate TABLE_PATH_TEMPLATE =
```
neozwu: It seems to me you need to configure resource name class for these.

igorbernstein2:
I would like to use static resource names. My only concern is that they would be duplicated & incompatible with cloud-bigtable-admin

neozwu:
@igorbernstein2 hmmm, my guess is If two proto resource classes with same name from bigtable and bigtable/admin/ were imported to the same java file, toolkit should be able to resolve it correctly.

igorbernstein2:
When I generated the admin, the TableName ended up as `com.google.bigtable.admin.v2.TableName`, I would imagine that enabling it here would generate it as `com.google.bigtable.v2.TableName`. Which I guess can be worked around by renaming the admin's package_name to `com.google.bigtable.v2`. But the we would run into the problem that you mentioned about the conflicting `package-info.java` file.

I'm starting to think that we should just keep the admin & data clients separate.

igorbernstein2:
In either case I created a CL to generate the type safe names

generated/java/gapic-google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/v2/BigtableClient.java
andreamlin: Blocked on cl/173174433.

andreamlin: Fixed.

andreamlin:
Updated the generated code with ResourceNames, reflecting the updated bigtable_gapic.yaml.
PTAL

PR #424, opened 2017-10-26
Refresh PHP

vam-google:


generated/php/google-longrunning-v1/tests/unit/GAX/LongRunning/OperationsClientTest.php
```diff
@@ -34,18 +34,27 @@
...
+use Google\Longrunning\DeleteOperationRequest;
+use Google\Longrunning\GetOperationRequest;
+use Google\Longrunning\ListOperationsRequest;
 use Google\Longrunning\ListOperationsResponse;
 use Google\Longrunning\Operation;
+use Google\Longrunning\OperationsGrpcClient;
 use Google\Protobuf\Any;
 use Google\Protobuf\GPBEmpty;
 use Grpc;
+use PHPUnit_Framework_TestCase;
```
garrettjonesgoogle: I see a lot of imports added here that aren't used... do we have no import cleanup step for PHP?

vam-google:
PTAL

generated/php/google-longrunning-v1/tests/unit/GAX/LongRunning/OperationsClientTest.php
vam-google: The issue should be fixed now (https://github.com/googleapis/artman/pull/310)

PR #425, opened 2017-10-26
Java: Regenerating for gax-httpjson

garrettjonesgoogle:


garrettjonesgoogle:
Note: Java will fail because gax hasn't been published yet at the referenced versions

andreamlin:
There's a diff in Spanner.proto. Does the version number have to be bumped?

garrettjonesgoogle:
Ugh that just kind of snuck in between regenerations. I'll revert that part. Good catch!

garrettjonesgoogle:
I regenerated with the toolkit update in the toolkit PR, and undid the spanner updates. PTAL

garrettjonesgoogle:
Updated with a couple more changes - sorry!

PR #432, opened 2017-11-05
Java: regenerating - updating resource name types

garrettjonesgoogle:
Also trimming api-common dependencies using https://github.com/googleapis/toolkit/pull/1669 .


pongad:
The changes look fine.
I might be missing something, but why do we check in proto files in this repo?

garrettjonesgoogle:
So we can see the diffs, and also we call "./gradlew uploadArchives" from this repo locally, so it's nice to have a commit for each release we do.

pongad:
LGTM

PR #437, opened 2017-11-07
Regenerate php bigtable, bigtable-admin using proto v3.4.0

andreamlin:
Generated with artman using protobuf/protoc 3.4.0.


andreamlin:
PTAL

PR #439, opened 2017-11-07
Java: promoting common-protos and language to 1.0.0

garrettjonesgoogle:


generated/java/grpc-google-longrunning-v1/build.gradle
```diff
@@ -10,7 +10,7 @@ apply plugin: 'signing'
 
 description = 'GRPC library for grpc-google-longrunning-v1'
 group = "com.google.api.grpc"
-version = "0.1.23"
+version = "0.1.24"
```
michaelbausor: Do we want to publish another version of this package? Or are we going to stop publishing it and make people use grpc-google-common-protos?

garrettjonesgoogle: I thought I would just publish this one last version and the remove it afterwards.

PR #467, opened 2017-11-21
Php GAPIC: Dlp v2beta1

ethanbao:
This pull request was generated by artman. Please review it thoroughly before merging.


michaelbausor:
This looks like it was generated with a config that is not from googleapis. Is that correct?

PR #468, opened 2017-11-22
Regenerate LongRunningClient with new namespace.

vam-google:
Update individual packages gax dependency.
Update (minor) DLP generated client.


vam-google:
@michaelbausor PTAL

Added videointelligence-v1
Regenerated bitable-admin-v1 under proper namespace

PR #472, opened 2017-12-05
Java batch update

andreamlin:


garrettjonesgoogle:
It looks like the build is failing because bigtable now needs iam.

garrettjonesgoogle:
Also you seem to be including class files, those should be removed.

generated/java/grpc-google-longrunning-v1/src/main/java/com/google/longrunning/OperationsGrpc.java
```diff
@@ -0,0 +1,552 @@
...
+ * operation resource, or pass the operation resource to another API (such as
+ * Google Cloud Pub/Sub API) to receive the response.  Any API service that
+ * returns long-running operations should implement the `Operations` interface
+ * so developers can have a consistent client experience.
+ * </pre>
+ */
+@javax.annotation.Generated(
+    value = "by gRPC proto compiler (version 1.2.0)",
+    comments = "Source: google/longrunning/operations.proto")
+public final class OperationsGrpc {
```
garrettjonesgoogle: Ok there is an issue - this shouldn't be added back. I think this is a mess I left behind. Basically, proto and grpc classes for longrunning are included in common-protos; we only want a gapic folder for longrunning.

andreamlin:
Removed class files, and fixed iam dependency in bigtable.

generated/java/grpc-google-longrunning-v1/src/main/java/com/google/longrunning/OperationsGrpc.java
andreamlin: ok, removed proto-google-longrunning-v1 and grpc-google-longrunning-v1

andreamlin: and replaced the longrunning gapic's dependency on those classes with the common-protos

andreamlin:
PTAL
Not sure why the openjdk7 build fails on a STANDARD_ERROR that passes in tests outside of the logging API; openjdk8 build does pass!

PR #475, opened 2017-12-06
Allow smoke tests to be present in PHP

michaelbausor:
- Allow smoke tests to be present by excluding them from normal test runs
- Add smoke test for pubsub
- Add system test config that can be used to run smoke tests

Smoke tests for other APIs can be added when those APIs are refreshed. This PR does not cause smoke tests to be run as part of api-client-staging CI, as I expect that this would cause too many failures at this point.


michaelbausor:
Yes, it means the smoke tests will be included in the repository, but that they will not be run, and that they will fail if you have not set the PROJECT_ID environment variable correctly.

PR #505, opened 2018-01-04
aggregate javadoc

pongad:
Introduce a new build target "allDocs" that creates Javadoc
for all proto- and grpc- subprojects.
This fixes a bug where grpc and proto docs weren't correctly linked
together.

We'll later be able to update all APIs in one git commit.

One caveat: all APIs will share the same version, so stable APIs
will have 0.1.x version.
If this is a serious problem, it can be fixed in a subsequent commit,
as the new target isn't used.


generated/java/version.txt
```diff
@@ -0,0 +1 @@
+0.1.27
```
garrettjonesgoogle: This could easily drift from the versions in the individual build.gradle files

generated/java/build.gradle
```diff
@@ -91,3 +91,17 @@ subprojects {
     exclude 'build/**'
   }
 }
+
+apply plugin: 'java'
+
+version = file("version.txt").text.trim()
+
+task allDocs(type: Javadoc) {
+  def subProjects = subprojects.findAll {it.name.startsWith("proto-") || it.name.startsWith("grpc-")}
```
garrettjonesgoogle: I can't quite tell what this does (or what it overrides).

generated/java/javadoc.sh
```diff
@@ -0,0 +1,30 @@
...
+
+api-client-staging: $REV
+"
+
+# Do this manually for now.
+# Can become fully automatic after we use the script a few times and have some more confidence in it.
+#   cd tmp_gh-pages
+#   git push
+#   cd ..
+#   rm -rf tmp_gh-pages
```
garrettjonesgoogle: Um uneasy about using shell scripts when we could put this into build.gradle...

pongad:
@garrettjonesgoogle Using Gradle instead now.

pongad:
@garrettjonesgoogle Using Gradle instead now.

generated/java/build.gradle
```diff
@@ -99,7 +99,62 @@ task allDocs(type: Javadoc) {
...
+    }
+
+    exec {
+      workingDir "tmp_gh-pages"
+      commandLine 'git', 'commit', '-m', "regenerate javadocs\n\napi-client-staging: ${gitHash}"
+    }
+  }
+}
+
+task pushDocs {
```
garrettjonesgoogle: Can we add removal of tmp_gh-pages to the clean task?

pongad:
@garrettjonesgoogle PTAL

generated/java/build.gradle
pongad: done

garrettjonesgoogle:
LGTM, but make sure at least one of the Java tests passes before merging

pongad:
@garrettjonesgoogle Fixed build, but I think you have to LGTM again.

PR #506, opened 2018-01-08
Java: add dlp v2beta2 

neozwu:
Add dlp v2beta2


generated/java/grpc-google-cloud-dlp-v2beta2/build.gradle
```diff
@@ -0,0 +1,167 @@
...
+  }
+}
+
+apply plugin: 'java'
+apply plugin: 'maven'
+apply plugin: 'signing'
+
+description = 'GRPC library for grpc-google-cloud-dlp-v2beta2'
+group = "com.google.api.grpc"
+version = "0.1.26"
```
garrettjonesgoogle: This has been updated to 0.1.27 in the googleapis config

neozwu:
PTAL

neozwu:
@garrettjonesgoogle @andreamlin @vam-google PTAL.  It seems openjdk7 has been failing before this change.

garrettjonesgoogle:
It appears that you regenerated using protoc 3.3; you need to use protoc 3.4. However, if you switch locally, the resource plugin will fail unless you also update your resource plugin to a more recent (unreleased) version which fixes the incompatibility problem with 3.4. It may just be easiest to wait until all the changes in flight have landed.

garrettjonesgoogle:
LGTM

PR #583, opened 2018-02-15

vam-google:
This PR is automatically generated by Artmanflow tool.

This does not include speech artman_speech_v1p1beta1


generated/java/grpc-google-longrunning-v1/build.gradle
```diff
@@ -0,0 +1,110 @@
+buildscript {
```
garrettjonesgoogle: You need to remove proto-google-longrunning-v1 and grpc-google-longrunning-v1

vam-google:
Please check https://github.com/googleapis/api-client-staging/pull/585 instead

PR #584, opened 2018-02-16
[DO NOT SUBMIT] Java regen 3: oneof -> inheritance 

garrettjonesgoogle:
This requires config changes in googleapis.

- parent abstract class using no prefix
- factory class (with the parent name + 's')
- deleting all deprecated stuff


garrettjonesgoogle:
This is a regen with the redesigned oneof -> inheritance conversion; it is not using the latest googleapis (I will have to do a rebase later).

michaelbausor:
A random sampling LGTM

PR #586, opened 2018-02-20
Java regen 4: oneof -> inheritance

garrettjonesgoogle:


garrettjonesgoogle:
PTAL - This should be the final version - the build just needs to be rerun after api-common 1.3.0 is released.

The third commit in the PR is a refresh with the latest googleapis.

PR #598, opened 2018-02-26
Add java batch script

neozwu:


generated/java/java_batch.py
```diff
@@ -0,0 +1,422 @@
...
+
+def fix_commons_proto_dep(flags):
+  file_list = _get_all_build_gradle_files(flags)
+  PROTO_PATTERN = 'compile project(":proto-google-longrunning-v1")'
+  PROTO_REPLACE = 'compile project(":proto-google-common-protos")'
+  GRPC_PATTERN = 'testCompile project(":grpc-google-longrunning-v1")'
+  GRPC_REPLACE = 'testCompile project(":grpc-google-common-protos")'
+  for f in file_list:
+    _fix_line_in_file(f, PROTO_PATTERN, PROTO_REPLACE)
+    _fix_line_in_file(f, GRPC_PATTERN, GRPC_REPLACE)
```
garrettjonesgoogle: Let's get this fixed in toolkit

generated/java/java_batch.py
```diff
@@ -0,0 +1,422 @@
...
+    'videointelligence_v1',
+    'vision_v1',
+    'vision_v1p1beta1',
+]
+
+# list of APIs which don't generate proto library
+PROTO_EXCLUSION = ['longrunning']
+
+# list of APIs which don't generate grpc library
+GRPC_EXCLUSION = ['appengine', 'longrunning']
```
garrettjonesgoogle: Let's get these exclusions fixed in artman config

generated/java/java_batch.py
```diff
@@ -0,0 +1,422 @@
...
+
+# list of APIs which `copy` subcommand doesn't copy from staging to gcj repo
+COPY_EXCLUSION = [
+        'speech_v1p1beta1',
+        'dlp_v2beta2',
+        'datastore',
+        'spanner_admin_database',
+        'spanner_admin_instance',
+        'spanner',
+]
```
garrettjonesgoogle: I would actually prefer the default to be no-copy, and then flip to copy. The reason is that when a new API is configured, there can be a period where we don't want to put the generated code into google-cloud-java yet; this scenario requires two flips (1. add to the exclusion list, 2. remove off the exclusion list). One way to make this easier to manage is to make the API list have attributes for whether the copy is on or not, so that we don't need two separate lists for proto/grpc generation vs copying.

generated/java/java_batch.py
```diff
@@ -0,0 +1,422 @@
...
+
+def get_artman_yaml(googleapis_repo):
```
garrettjonesgoogle: It seems like this function should be pluralized.

generated/java/java_batch.py
```diff
@@ -0,0 +1,422 @@
...
+  return [
+      yaml for yaml in artman_yaml_files
+      if os.path.basename(yaml) not in exclusion_list
+  ]
+
+
+def api_to_yaml_mapping(artman_yaml_files):
+  mappings = dict(
+      zip([
+          os.path.basename(yaml_file)[7:-5] for yaml_file in artman_yaml_files
```
garrettjonesgoogle: These numbers are too magick-y; could we do substring comparisons instead?

neozwu:
Close this PR since it's irrelevant now :(

PR #605, opened 2018-03-05
Bump grpc/proto versions

andreamlin:


generated/java/proto-google-cloud-pubsub-v1/build.gradle
```diff
@@ -10,7 +10,7 @@ apply plugin: 'signing'
 
 description = 'PROTO library for proto-google-cloud-pubsub-v1'
 group = "com.google.api.grpc"
-version = "0.2.1"
+version = "0.3.0"
```
garrettjonesgoogle: Shouldn't the api-common dependency change below? Did you generate based on your update to googleapis?

andreamlin: Ahh I will regenerate everything ok. This PR was just a sed job.

andreamlin:
Regenerated the Java packages with the latest googleapis. PTAL

andreamlin:
Actually not ready yet. Not all api-common versions got bumped.

andreamlin:
Ok, with the exception of dlp-v2 which is in googleapis-private, all api-common dependencies are bumped to 1.4.0. PTAL

andreamlin:
PTAL

PR #606, opened 2018-03-06
Bump versions for dlpv2 too

andreamlin:
Grpc/proto package version for dlpv2 didn't get updated along with all the other APIs in the past commit.


andreamlin:
PTAL

andreamlin:
@michaelbausor Thanks. There isn't a googleapis/ artman config file for dlp-v2, so I didn't regenerate it. @neozwu has more details.

PR #612, opened 2018-03-15
add video v1p1beta1 and batch refresh

neozwu:


andreamlin:
For future version bump PRs, it would be nice to have the batch refreshed *.java and *.proto diffs in one PR, and then the build.gradle version bump in a separate PR.

PR #613, opened 2018-03-15
Regenerate bigtable

igorbernstein2:
* Add resource headers
* Mark CheckConsistency RPC as idempotent

The googleapis gapic changes were sync in this commit:
https://github.com/googleapis/googleapis/commit/de4a5ff6534892029e083d7c0651e06e98c51729


generated/java/proto-google-cloud-bigtable-admin-v2/src/main/java/com/google/bigtable/admin/v2/AppProfileName.java
```diff
@@ -112,25 +109,6 @@ public static boolean isParsableFrom(String formattedString) {
     return PATH_TEMPLATE.matches(formattedString);
   }
 
-  public Map<String, String> getFieldValuesMap() {
```
garrettjonesgoogle: This should be retained

PR #614, opened 2018-03-19
Staging cleanup, Artman Workflow Java sources staging 

vam-google:
This PR is a result of integrating all of the pending java generation cleanup changes:
- https://github.com/googleapis/artman/pull/400
- https://github.com/googleapis/toolkit/pull/1867
- pending googleapis (artman_*.yaml) cleanup change

Also:
- Publish recent bigtable api changes.
- Upgrade root `gradlew` version from `2.11-bin` to `4.0.1-all` (same as in `toolkit`).
- Upgrade `junit` dependency version from `4.11` to `4.12` (same as `google-cloud-java`)
- Remove maven publishing steps from generated `proto/grpc build.gradle` files to root `build.gradle` (the generated artifacts should not imply only one way of publishing, plus this makes the generator output significantly shorter).
- Cleanup root and generated build files (remove unused jacoco build targets, remove unused ext properties in generated file, remove unused repositories); removing jacoco build target makes build+test time ~2 times faster.
- Make generated output self-contained (it will still need a root `build.gradle` file, but the most basic one is sufficient (just to be able to run gradle from root directory). 
- Cleanup and simplify logic for pom creation: now `./gradlew install` and `./gradlew uploadArchives` will create exactly same pom artifact (they used to be differrent, local did not include any custom information added for uploadArchives). Pom information was cleaned up too: corrected scm info, removed developer info (developer info, since it was about company, not people, which is incorrect).
- Gax and gax-grpc dependency versions are now determined on generation step (how  it is supposed to be by initial design), instead of being determined by staging root build.gradle


garrettjonesgoogle:
LGTM once the build issues are sorted out

vam-google:
The build is fixed, googleapis, artman and toolkit changes are all merged. @garrettjonesgoogle PTAL

PR #616, opened 2018-03-20

michaelbausor:
This PR is automatically generated by Artmanflow tool


generated/java/proto-google-iam-v1/build.gradle
```diff
@@ -22,7 +22,7 @@ repositories {
 dependencies {
   compile "com.google.protobuf:protobuf-java:3.4.0"
   compile "com.google.api:api-common:1.4.0"
-  compile project(":proto-google-common-protos")
+  compile "com.google.api.grpc:proto-google-common-protos:1.4.0"
```
garrettjonesgoogle: This line needs to be reverted (we don't currently generate the right thing here)

PR #646, opened 2018-04-26

garrettjonesgoogle:
This PR is automatically generated by Artmanflow tool


garrettjonesgoogle:
According to @michaelbausor the php failures are expected since the build pulls in the master branch of gax-php, which requires an update in generated php code, which isn't reflected in the repo yet.

PR #650, opened 2018-05-14
add resource name types

pongad:


pongad:
We decided to not do this.

PR #666, opened 2019-02-05
Regen Java proto and grpc common protos

andreamlin:
Note: the master branch currently contains the protos from api-common-protos/input-contract, which is contains a slightly different set of the protos currently used by google-cloud-java. This explains why the diff is so large, and why entities like `Metadata` are removed (they only exist in api-common-protos/input-contract) and why entities like google/api/experimental are included (they still exist in 1.12.0)

Steps to create this:
1. Assume cl/232550844 is checked in; this excludes googleapis:api/expr from the input set of protos to be generated
2. In googleapis/master: run `artman --config google/artman_core.yaml generate java_grpc`
3. In api-client-staging, remove the existing `generated/java/grpc-google-common-protos` and `generated/java/proto-google-common-protos` directories. Move the artman-generated directories to replace the above deleted directories.
4. Remove the generated pom.xmls. Update the generated build.gradle files with the new version 1.14.0.


andreamlin:
openjdk7 test has been failing for a while. something about networks and ssl maybe.

andreamlin:
cc @kolea2

andreamlin:
PTAL

generated/java/proto-google-common-protos/src/main/java/com/google/api/ExperimentalProto.java
```diff
@@ -1,10 +1,10 @@
 // Generated by the protocol buffer compiler.  DO NOT EDIT!
-// source: google/api/signature.proto
+// source: google/api/experimental/experimental.proto
```
michaelbausor: Do we want to include the protos in the experimental directory?

generated/java/proto-google-common-protos/src/main/java/com/google/api/AnnotationsProto.java
```diff
@@ -7,17 +7,6 @@
   private AnnotationsProto() {}
   public static void registerAllExtensions(
       com.google.protobuf.ExtensionRegistryLite registry) {
-    registry.add(com.google.api.AnnotationsProto.metadata);
```
michaelbausor: It looks like lots of stuff is being pulled out of annotations.proto. Is that what we expect to happen?

PR #661, opened 2018-08-22
Regenerate common protos from api-common-protos/input-contract

andreamlin:
This PR is needed to bootstrap the support of proto annotations in gapic-generator. gapic-generator requires the protos from https://github.com/googleapis/api-common-protos/tree/input-contract

Steps to create this PR:
- Checked out [googleapis/api-common-protos:input-contract](https://github.com/googleapis/api- common-protos/tree/input-contract) locally.
- Copied over artman_core.yaml from googleapis/googleapis into the local api-common-protos.
- Ran `artman --config google/artman_core.yaml java_grpc`
- In api-client-staging, deleted exisiting grpc- and proto-google-common-protos, and replaced them with the artman-generated directories.
- Removed pom.xmls.
- Updated version to 0.12.0-beta (from 0.12.0)

PR #667, opened 2019-02-20
Regen input contract

andreamlin:
Regenerating api-common:input-contract protos for Java.
Followed the same steps in https://github.com/googleapis/api-client-staging/pull/661.

Note: Moved all the input-contract protos to a separate branch on this repo called input-contract. (It gets confusing when we have a PR for common-protos at master against common-protos at the input-contract state and then again vice versa.) This means that this PR's diff is only against the newest version of the input-contract protos and the last Java-published version of the input-contract protos.


michaelbausor:
Why are there changes to python files?

andreamlin:
Oh my, I don't know how the python changes snuck in. Reverted those.

PR #671, opened 2019-07-03
regen iam

noahdietz:
Fixes https://github.com/googleapis/gapic-generator/issues/2865 and other issues

CC: @crwilcox


noahdietz:
Per offline discussion with @lukesneeringer, the moving of the grpc code to the new files is not a concern, it shouldn't have been used anyways. I will regen with the [protoc-docs-plugin](https://github.com/googleapis/protoc-docs-plugin) and it should be good-to-go. Furthermore, CI can be ignored here.

noahdietz:
Pinning the `nox` version fixed the python-related CI. Moving forward with this change.

PR #672, opened 2019-07-09
Releasing IAM v0.12.0

alexander-fenster:
Includes the code generated from the new IAM protos: #671

### DIFF
--- a/generated/python/grpc-google-iam-v1/setup.py
+++ b/generated/python/grpc-google-iam-v1/setup.py
@@ -16,7 +16,7 @@ install_requires = [
 
 setuptools.setup(
   name='grpc-google-iam-v1',
-  version='0.11.4',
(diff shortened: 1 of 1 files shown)


alexander-fenster:
@crwilcox are you ok with us making this release now? Just to confirm. Thanks!

noahdietz:
Discussed offline, we can release this. It won't actually go into affect until we update the generated dependency version in [gapic-generator](https://github.com/googleapis/gapic-generator/blob/f4f33321136f930bbe6b3a60208796e7380f9453/src/main/resources/com/google/api/codegen/packaging/dependencies.yaml#L88), so we will see this fail in CI if it still doesn't work. It shouldn't affect existing clients (unless we released a `0.12dev`, the current upper-bound).

PR #682, opened 2020-09-19
chore: add codeowners and repo settings

JustinBeckwith:


hkdevandla:
This repo doesn't look like up-to-date or used actively based on the commits/generated test libraries. Adding languages owners to chime in if we need to adjust/fix the required checks.

@vam-google @alexander-fenster @miraleung @software-dov

vam-google (commented):
@hkdevandla This repo has been occasionally used to publish some of the core libraries (at least for java) because there was no alternative way of doing it at that moment. I believe it was specifically about publishing java common proto maven artifact and java iam maven artifact. If Yoshi now has a way to publish those in a different way then, I think, this repo may be deprecated.

As for failing PHP targets, if I remember correctly they have been failing for a really long time, even before this repo became obsolete. They are "safe" to ignore.

hkdevandla:
Based on discussion with folks, this repo seems to be the only source for publishing appengine artifacts, so I would consider to keep this repo in place. We will clean up the directories later for other non-java languages (as they are not used anymore).

.github/sync-repo-settings.yaml
```diff
@@ -4,16 +4,10 @@ mergeCommitAllowed: false
...
-    - 'ci/circleci: openjdk7'
-    - 'ci/circleci: php55'
-    - 'ci/circleci: php56'
-    - 'ci/circleci: php70'
-    - 'ci/circleci: php71'
     - 'ci/circleci: node4'
     - 'ci/circleci: node6'
     - 'ci/circleci: node7'
     - 'ci/circleci: node8'
-    - 'ci/circleci: openjdk8'
```
hkdevandla: should we consider keeping jdk8 since the tests for that are passing?

hkdevandla (approved):
Added a comment about circleci for jdk8. Otherwise looks good. Thanks!

